PR #1, opened 2019-06-14
Add compatibility for 20% of community

arne182:


Gernby:
It seems that there is a mixture of case conventions.  It seems we should all use camel case here.

arne182:
I agree

geohot:
Merging is not a democracy. These changes have no case convention and no comments. What's a "readdistancelines"? "lkMode"? Do "distanceToggle", "laneDepartureToggle", etc.. not belong in ButtonState?

Do these things really belong in CarState? If so, they should be much better commented and understandable, compare to what's already there.

arne182:
This is what the community has added over the past year from @joeljacobs to  @kegman and every body inbetween. Moving them out into their own capnp was even suggested.

PR #2, opened 2019-06-14
Add deadzone parameters for pathplanner

Gernby:
This is related to another PR I'll be submitting later today.


geohot:
So I looked at the other PR, it has cereal changes besides this. Also, comments and more description for me to be able to intelligently respond! "This is related to another PR I'll be submitting later today." isn't a description.

Gernby:
Thanks for the response.  I actually need to close this PR, since I closed the related PR.  However, I'm still a bit unsure about the purpose or goal of submitting cereal parameters separately in this repo.  It seems that the only benefit of this being separate would be to improve Comma's data processing from custom forks.  If so, then it seems that the requirements for new parameters should be pretty light, with quick turn-around time.  If a new parameter requires detailed justification with supporting data, then it seems that this repo just increases effort for everyone.

PR #3, opened 2019-07-06
Differentiate transmission types

csouers:
We have some users with manual trans Hondas that I'd like to work on getting more upstreamable. This is part 1 of that process. I did not include a line for EVs, as that could be considered part of 'automatic'.

Ran make with no errors.

### DIFF
--- a/car.capnp
+++ b/car.capnp
@@ -269,7 +269,6 @@ struct CarControl {
       wrongGear @4;
       seatbeltUnbuckled @5;
       speedTooHigh @6;
-      ldw @7;
     }
 
     enum AudibleAlert {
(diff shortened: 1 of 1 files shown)


jyoung8607:
I ask for this PR, or something like it, to be merged in support of the Volkswagen community port. We are asking for it as part of the work in #8:

> We already have manual transmission users on the community Volkswagen port, and right now we're having to pretend they're in Drive all the time. We'd like to give them better safety handling, and surely your upload and ingest pipeline will appreciate not being lied to... if nothing else, this will give you a handy marker to identify them and toss the data if you don't want it!

rbiasini:
merged after fixing conflicts. Thanks!
https://github.com/commaai/cereal/commit/24f6fa78cdd0e357535b70328465c3c7003bb921

PR #5, opened 2019-07-25
Add lane departure warning to visual hud alerts

csouers:
A follow up PR for openpilot and opendbc (for Honda maybe with a refactor for Toyota) will be coming soon.


arne182:
For Toyota the ldw is already on the visual hud alerts

PR #7, opened 2019-09-24
Add this for future Blindspot recognition

arne182:
I am not sure if a boolean is good instead of sending the distance to object but it would be interesting to talk about this.

### DIFF
--- a/car.capnp
+++ b/car.capnp
@@ -182,6 +182,7 @@ struct CarState {
       altButton3 @8;
     }
   }
+  blindspot @27 :Bool; # Is there something blocking the lane change
 }
 
 # ******* radar state @ 20hz *******


car.capnp
```diff
@@ -182,6 +182,7 @@ struct CarState {
       altButton3 @8;
     }
   }
+  blindspot @27 :Bool; # Is there something blocking the lane change
```
rbiasini: I think at least distinguishing between left VS right should be required.
Also, maybe anything that has to do with perception should be handled by radard. I'll think about this.

arne182:
I also thought that having both is good in the beginning. But if we limit the auto Lane change based on blinker on direction. You should only have a violation in that direction. Checking both sides requires more code on Toyota and does not give more valuable data. Later it could be used for crash avoidance. But that is another topic.

geohot:
Is this being used in forks? I'd be fine with upstreaming blindspot stuff.

arne182:
We are working on it currently

arne182:
Will reopen a PR to be up to date with master.

PR #11, opened 2019-11-07
Changes to regex so js bindings can be generated correctly

chrisinajar:


geohot:
Is this still relevant?

PR #12, opened 2019-11-19
Switch from polling on FIFOs to signal

geohot:
First off, remove randomness unless there's a real justification for it. In general, unless you absolutely need it, randomness shouldn't be used.

Secondly, with uids being pids, I think the signalling solution is obvious. The polling process can sleep until the publishing process decides it's time to wake it up. I don't know the nuance here, but usleep can be exited by EINTR, though I think the signal handler might have to be configured to do that.

You could get more fine grained with the poll, where the subscriber says if it's blocked on a given socket to avoid spurious wakeups (right now it wakes on all subscribed sockets), but this is a perf optimization for later and maybe not needed at all.

Right now this segfaults with python demo.py, and I don't know why. In general, this software needs a much better test suite, it underlies all of system stability and has to deal with many different orders of events.

### DIFF
--- a/messaging/msgq.cc
+++ b/messaging/msgq.cc
@@ -414,18 +414,16 @@ int msgq_poll(msgq_pollitem_t * items, size_t nitems, int timeout){
     if (items[i].revents) num++;
   }
 
+  int ms = (timeout == -1) ? 100 : timeout;
+  struct timespec ts;
+  ts.tv_sec = ms / 1000;
+  ts.tv_nsec = (ms % 1000) * 1000 * 1000;
+
 
   while (num == 0) {
-    // TODO: if a message is ready on any of the sockets, don't sleep
-    // TODO: switch to nanosleep and store remaining time in case there is a false positive
-
     int ret;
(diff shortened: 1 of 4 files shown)


messaging/msgq.cc
```diff
@@ -465,36 +388,24 @@ int msgq_msg_recv(msgq_msg_t * msg, msgq_queue_t * q){
...
-    if (fds[i].revents && !items[i].revents){
-      // Don't add it if it was already added
-      num++;
-      items[i].revents = 1;
+    // get number to return
+    for (size_t i = 0; i < nitems; i++) {
+      num += msgq_msg_ready(items[i].q);
     }
+
+    // exit if we had a timeout and the sleep finished
```
geohot: hmm, we have to track the time passed.

pd0wm:
Sounds like nanosleep is the way to go here. It is designed to be interrupted, has nice exit codes and even stores the remaining time when interrupted in a new struct.

http://man7.org/linux/man-pages/man2/nanosleep.2.html

PR #15, opened 2019-11-22
Remove ZMQ publish socket assert() from messaging layer

jyoung8607:
# Bug Fix

This duplicates PR commaai/openpilot#884 in the new home of the messaging library to make sure the fix is preserved beyond OP 0.6.6.

### DIFF
--- a/messaging/impl_zmq.cc
+++ b/messaging/impl_zmq.cc
@@ -111,7 +111,7 @@ void ZMQPubSocket::connect(Context *context, std::string endpoint){
 
   //std::cout << "ZMQ PUB: " << full_endpoint << std::endl;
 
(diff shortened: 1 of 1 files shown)


pd0wm:
You want the PubSocket create to fail if it can't bind to the socket (e.g. if there is already another publisher). The error handling needs be better and it needs to be communicated to the caller instead of crashing, but we can't just silently fail.

jyoung8607:
> You want the PubSocket create to fail if it can't bind to the socket (e.g. if there is already another publisher). The error handling needs be better and it needs to be communicated to the caller instead of crashing, but we can't just silently fail.

We're in agreement there (see my opening comment on the other PR) but in the meantime I needed a solution for openpilot-tools to work. You've provided that (thanks again!) so I'll close this.

PR #25, opened 2020-02-02
GitHub actions

quillford:
Moved CI over to github actions so the output can be seen by external contributors. With azure pipelines, it's just pass/fail. Also, forks get CI by default.


geohot:
@pd0wm This good?

PR #29, opened 2020-02-14
add honda ecus

jpancotti:
add honda ecu names

### DIFF
--- a/car.capnp
+++ b/car.capnp
@@ -468,10 +468,21 @@ struct CarParams {
     fwdCamera @3;
     engine @4;
     unknown @5;
+    transmission @8; # Transmission Control Module
+    srs @9; # airbag
+    gateway @10; # can gateway
+    hud @11; # heads up display
+    combinationMeter @12; # instrument cluster
 
     # Toyota only
     dsu @6;
     apgs @7;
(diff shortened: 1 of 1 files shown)


pd0wm:
Maybe we should use more descriptive names in the capnp than honda uses. Think we can keep VSA and SRS since those are common abbreviations. 

eg:
```
    vsa @8; # Vehicle Stability Assist
    programmedFuelInjection @9;
    shiftByWire @10;
    transmission @11; # Transmission Control Module
    srs @12; # airbag
    electricBrakeBooster @13;
    combinationMeter @14;
    gateway @15;
    hud @16; # heads up display
```

We can also move some of them into the common section since they don't sound very honda specific like: gateway. transmission, srs and hud.

pd0wm:
You can't change the numbering for the existing enum values, that would make the old logs unreadable. Can you make sure numbering for those stays the same? The definition doesn't have to be in order of the numbers.

jpancotti:
> You can't change the numbering for the existing enum values, that would make the old logs unreadable. Can you make sure numbering for those stays the same? The definition doesn't have to be in order of the numbers.

fixed

PR #32, opened 2020-02-28
Add laneChangeBlocked to LaneChangeState

arne182:
For blind spot sensor data add the lane change state for lane change not possible

### DIFF
--- a/log.capnp
+++ b/log.capnp
@@ -779,9 +779,9 @@ struct PathPlan {
   enum LaneChangeState {
     off @0;
     preLaneChange @1;
-    laneChangeBlocked @2;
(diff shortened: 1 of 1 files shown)


geohot:
I wouldn't change the existing numbering

pd0wm:
Not sure if we need an extra state for this. I would think it's the same as preLaneChange, but it just wouldn't go into the next state.

I'll leave this PR open until we get around to implementing it in openpilot and we know what we need.

arne182:
I have also implemented something just using the preLaneChange and the CS.blindspotright. it is enough information if we can get the car cereal values integrated

PR #33, opened 2020-03-05
Add `wait_for` input parameter to SubMaster

ShaneSmiskol:
1. Add input parameter `wait_for` which takes a list of services to wait for when updating with the SubMaster.
   Example:
   ```python
   sm = messaging.SubMaster(['testStruct', 'thermal'], wait_for=['testStruct'])
   sm.update(0)  #  waits until new message from testStruct is received
   print('Updated: {}'.format(sm.updated['testStruct']))  # this should always return True
   ```

@pd0wm I tried what you suggested, only adding the poller to the services to wait for but that resulted in completely skipping those services if there wasn't a new message waiting to be received. So I did the opposite which is what I think you meant to say; for services to wait for, I did not specify a poller and instead in the update function of SM, I used `recv_one` instead of `recv_one_or_none`. This also required changing the `sock` class variable to include a `wait` flag so I could target those sockets in the update function after the fact.

No performance hit should be taken if there are no messages to be waiting for, other than iterating through the dictionary checking for the `wait` key. This could also be skipped by adding an if check above the `for service in self.sock:` loop checking if `self.wait_for` is None.

### DIFF
--- a/messaging/__init__.py
+++ b/messaging/__init__.py
@@ -125,7 +125,7 @@ def get_one_can(logcan):
       return can
 
 class SubMaster():
-  def __init__(self, services, ignore_alive=None, addr="127.0.0.1"):
+  def __init__(self, services, wait_for=None, ignore_alive=None, addr="127.0.0.1"):
     self.poller = Poller()
     self.frame = -1
     self.updated = {s : False for s in services}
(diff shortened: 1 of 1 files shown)


ShaneSmiskol:
I'd be happy to open a PR for master, converting everything to this new standard if approved, just let me know.

pd0wm:
I meant the first thing, where you put the poller on things you want to wait for. You then do a non blocking receive on the remaining sockets after.

What feature needs these changes? I don't see how this makes existing code shorter or more readable.

I like the Pubmaster.new_message() function. Can you put that in a separate PR, that sound like a no-brainer.

geohot:
I don't think new_message belongs in pubmaster, just messaging.

pd0wm:
I agree. Add the union type as an argument to the already existing `messaging.new_message()` function.

PR #35, opened 2020-03-05
Create and init message in one line with `new_message` in messaging 

ShaneSmiskol:
`new_message()` includes service and optional list size parameter to create a new message in just 1 line.

   Old code:
   ```python
   pm = messaging.PubMaster(['carControl'])
   cc_send = messaging.new_message()
   cc_send.init('carControl')
   cc_send.valid = CS.canValid
   cc_send.carControl = CC
   pm.send('carControl', cc_send)
   ```
   New code:
   ```python
   pm = messaging.PubMaster(['carControl'])
   cc_send = messaging.new_message('carControl')
   cc_send.valid = CS.canValid
   cc_send.carControl = CC
   pm.send('carControl', cc_send)
   ```

### DIFF
--- a/messaging/__init__.py
+++ b/messaging/__init__.py
@@ -19,10 +19,14 @@ except ImportError:
 
 context = Context()
 
-def new_message():
+def new_message(service, size=None):
   dat = log.Event.new_message()
   dat.logMonoTime = int(sec_since_boot() * 1e9)
   dat.valid = True
+  if size is None:
+    dat.init(service)
+  else:
+    dat.init(service, size)
   return dat
 
 def pub_sock(endpoint):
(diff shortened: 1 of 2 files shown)


pd0wm:
Can you make service an optional argument? By doing it that way all the existing code keeps working.

ShaneSmiskol:
Yes sir, just noticed that one of the calls didn't init, so I added just that

messaging/__init__.py
```diff
@@ -19,10 +19,14 @@
 
 context = Context()
 
-def new_message():
+def new_message(service=None, size=None):
   dat = log.Event.new_message()
   dat.logMonoTime = int(sec_since_boot() * 1e9)
   dat.valid = True
+  if service is not None and size is None:
```
pd0wm: Can you do a nested if here? Think that would me more easy to read.

```python
if service is not None:
  if size is None:
    dat.init(service)
  else:
    data.init(service, size)
```

ShaneSmiskol: Sure, let me just push the commit

PR #38, opened 2020-03-16
Additional car params auto-detection in support of VW

jyoung8607:
Allow auto-detection of some important params in preparation for decent VW fingerprinting. These are things we need to discover up-front in get_params() and preserve for later use.

- Add Enum for installed network location, so we can handle folks wired to the camera (R242) vs the gateway (J533). I know Comma likes standard camera integration, but gateway integrations are a popular, growing, and essential "fact on the ground". All we need is a CarParams flag so that when we need to, we can relocate one (1) CarState signal and retarget one (1) CarController can_sends.

- Add an electric vehicle variant to the transmission type Enum. Turns out for true EVs (not hybrids) we have to learn the gearshift lever position from an EV-specific signal.

### DIFF
--- a/car.capnp
+++ b/car.capnp
@@ -459,7 +459,7 @@ struct CarParams {
     unknown @0;
     automatic @1;  # Traditional auto, including DSG
     manual @2;  # True "stick shift" only
-    none @3;  # Electric vehicle or other direct drive
+    direct @3;  # Electric vehicle or other direct drive
   }
 
   struct CarFw {
--- a/car.capnp
+++ b/car.capnp
@@ -457,8 +457,9 @@ struct CarParams {
 
   enum TransmissionType {
     unknown @0;
-    automatic @1;
-    manual @2;
+    automatic @1;  # Traditional auto, including DSG
+    manual @2;  # True "stick shift" only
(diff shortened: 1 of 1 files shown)


car.capnp
```diff
@@ -456,8 +457,9 @@ struct CarParams {
 
   enum TransmissionType {
     unknown @0;
-    automatic @1;
-    manual @2;
+    automatic @1;  # Traditional auto, including DSG
+    manual @2;  # True "stick shift" only
+    direct @3;  # Electric vehicle or other direct drive
```
rbiasini: is the distinction between "direct" and "automatic" relevant in the upcoming openpilot changes? If not, I think automatic could just include "direct" transmissions too.

jyoung8607: Yes, that's why I need it. For auto MQBs except e-Golf, including hybrids, we get gearshift position from a message sent by the auto transmission controller. For the e-Golf, there isn't one, so we have to fish it out of an EV-specific message.

e-Golf: EV_Gearshift.GearPosition
Manual trans: ( synthesized from ECM clutch-pressed switch and BCM reverse light )
Everything else: Getriebe_11.GE_Fahrstufe

PR #42, opened 2020-05-09
Sub pub master

deanlee:
reusing an kj::array to do alignment related copy.
poll() will validate the incoming message,and return an event array
pollRaw() and receive() will not read the event. caller must call getEvent to parse and get the event.
check if message is "alive" when the optional parameter "alive" is set to true.but I have no idea where to use this flag.so I keep the message processing mode remains the same as before.: commaai/openpilot#1476
@pd0wm @adeebshihadeh

### DIFF
--- a/messaging/socketmaster.cc
+++ b/messaging/socketmaster.cc
@@ -1,0 +1,160 @@
+#include <assert.h>
+#include "messaging.hpp"
+#include "services.h"
+
+SubMaster::SubMaster(Context *context) : ctx(context), poller(nullptr), ownContext(false), frame(0) {
+  if (ctx == NULL) {
+    ctx = Context::create();
+    ownContext = true;
+  }
+  poller = Poller::create();
+}
+
+SubMaster::SubMaster(std::vector<const char *> services, const char *address, bool conflate, Context *context) : SubMaster(context) {
+  for (auto name : services) {
+    createSocket(name, address, conflate);
+  }
+}
+
+SubMaster::~SubMaster() {
+  delete poller;
+  for (auto s : sockets) {
+    delete s.second;
+    delete s.first;
+  }
+  if (ownContext) delete ctx;
+}
+
+SubMessage *SubMaster::createSocket(const char *endpoint, const char *address, bool conflate) {
+  SubMessage *msg = NULL;
+  for (const auto &it : services) {
+    if (strcmp(it.name, endpoint) == 0) {
+      SubSocket *socket = SubSocket::create(ctx, endpoint, address ? address : "127.0.0.1", conflate);
+      assert(socket != NULL);
+      poller->registerSocket(socket);
+      msg = new SubMessage(it.name, it.frequency, it.decimation);
+      sockets[socket] = msg;
+      break;
+    }
+  }
+  assert(msg != NULL);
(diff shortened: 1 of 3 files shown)


deanlee:
I have tested it and it seems to work fine.

pd0wm:
I'm confused by all the extra inputs and classes needed. Can we make the API exactly the same as the python version? Would be nice we can use the [] operator overloading to get the ` cereal::Event::Reader ` out.

deanlee:
Although I have tried to keep the code as simple as possible， but u r right, we can make the interface clearer and easier to use,I will continue to optimize it.

pd0wm:
I'm fine with loggerd not using submaster since it's use case is so different.

deanlee:
**Examples**
single sub without poller:
```c++
 SubMessage msg(NULL, "sensorEvents");
 cereal::Event::Reader *r = msg.receive();
 if (r) printf("%d", r->getValid());

```
with poller:
 ```c++
SubMaster sm(NULL, {"controlsState", "sensorEvents", "cameraOdometry"});
 sm.update(1000);
 for (cereal::Event::Reader event : sm.allAliveAndValid()){}

```
single pub:
```c++
 PubMessage p(NULL, "sensorEvents");
 p.send(msg);
```
multiple pub:
```c++
 PubMaster p(NULL, {"controlsState", "sensorEvents", "cameraOdometry"});
 p.send("controlsState", msg);
```

pd0wm:
Do we need the extra SubMessage class? I would just put it all in SubMaster. We don't need to add complexity for single socket subscribers.

The all_alive_and_valid method in python returns a bool, not a list of sockets. 

Can you overload the `[]` operator to get messages from the SubMaster?

like
```c++
SubMaster sm( {"controlsState", "sensorEvents", "cameraOdometry"});
sm.update(1000);

if (sm.updated["controlsState"]) {
  cereal::Event::Reader e = sm["controlsState"];
  // do stuff
}
```

deanlee:
updated.all in SubMaster now.
the only difference is using sm.updated("controlsState") instead of sm.update["controlsState"]. 

to support "updated[]",I  need add a new public member variable std::map. do we need that?

pd0wm:
@deanlee Thanks! That looks like the API I had in mind. Updated as a function is fine!

It's a big change, so I'll need some time to review it. Can you make sure your openpilot PR is properly rebased on latest master? Is see some unrelated changes to stuff like `read_db_value` in there.

deanlee:
my pleasure.I have rebased it to latest master branch.

messaging/socketmaster.cc
```diff
@@ -0,0 +1,162 @@
...
+    SubMessage *m = messages_.at(s);
+    const size_t size = (msg->getSize() / sizeof(capnp::word)) + 1;
+    if (m->buf.size() < size) {
+      m->buf = kj::heapArray<capnp::word>(size);
+    }
+    memcpy(m->buf.begin(), msg->getData(), msg->getSize());
+    delete msg;
+
+    if (m->msg_reader) {
+      m->msg_reader->~FlatArrayMessageReader();
```
pd0wm, line 86: why not `delete m->msg_reader`?

PR #47, opened 2020-05-22
Custom Messagebuilder

deanlee:
Stack based MessageBuilder, saves 2 malloc and one copy.

usage:

 ```c++
MessageBuilder msg;
auto framed = msg.initEvent().initFrame();
...
pm.send("frame", msg);
```
openpilot PR: https://github.com/commaai/openpilot/pull/1555

### DIFF
--- a/messaging/messaging.hpp
+++ b/messaging/messaging.hpp
@@ -1,6 +1,7 @@
 #pragma once
 #include <cstddef>
 #include <map>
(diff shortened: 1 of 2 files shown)


pd0wm:
Is there really a benefit over using `MallocMessageBuilder`? There must be significant CPU saving for it to be worth the complexity.

PR #50, opened 2020-06-06
fix mac build

geohot:
It seems to me like messaging/socketmaster.cc was in the wrong place in the build script.


geohot:
@pd0wm I see you merged #42, was there a reason for this?

deanlee:
it seems that offroad.apk will fail to start if put socketmaster.cc in messaging_objects.

geohot:
Hmm, sounds like we have to fix offroad. It really doesn't belong in cereal.

PR #61, opened 2020-07-05
k-line 5 init fault type

gregjhogan:


gregjhogan:
https://github.com/commaai/panda/pull/565

PR #57, opened 2020-06-16
Add class SubMessage

deanlee:
Add class SubMessage to handle single SubSocket. refactor SubMaster to using SubMessage internally

**non conflated**
```c++
SubMessage msg( "sensorEvents");
if (msg.receive()) {
  cereal::Event::Reader &r = msg.getEvent();
}
```

**conflated**
```c++
SubMessage msg( "sensorEvents", "127.0.0.1", true);
if (msg.receive()) {
  cereal::Event::Reader &r = msg.getEvent();
}
```

PR #69, opened 2020-07-27
SubMaster refactor

adeebshihadeh:
* [x] refactor PubMaster
* [ ] refactor SubMaster
  * [ ] python and c++ API match
* [ ] unit tests for SubMaster validity and alive check
* [ ] merge in #57


adeebshihadeh:
This doesn't actually save any CPU, and we'd have to maintain a completely different submaster for testing.

PR #77, opened 2020-08-03
Add support for Volvo

danielzmod:
Add support for Volvo V40 and placeholder for EUCD-platform vehicles.


adeebshihadeh:
How different are these two platforms? Do they really require two different safety modes?

danielzmod:
> How different are these two platforms? Do they really require two different safety modes?

They have totally different message adresses. But they could be separated easily by providing a parameter. Or do some checks for specific messages.

danielzmod:
@adeebshihadeh Do you want me to merge to one instead?

PR #84, opened 2020-08-31
allow prioritization of services in SubMaster

adeebshihadeh:
This fixes high CPU usage in processes like dmonitoringd which should run at 10hz driven by the driverState packet, but it's main loop runs at 100hz because it subscribes to carState. Blocking on `driverState` reduces dmonitoringd's CPU usage from 6% to 2%.

### DIFF
--- a/messaging/__init__.py
+++ b/messaging/__init__.py
@@ -149,7 +149,7 @@ class SubMaster():
     for s in services:
       if addr is not None:
         p = self.poller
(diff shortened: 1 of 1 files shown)


pd0wm:
Why keep two pollers around? It might be simpler to just have one poller, and then call a non blocking receive on all the sockets in the SubMaster.

PR #87, opened 2020-09-11
fix the isssue 41

grekiki:
https://github.com/commaai/cereal/issues/41

### DIFF
--- a/messaging/msgq.cc
+++ b/messaging/msgq.cc
@@ -21,6 +21,8 @@
 
 #include <stdio.h>
 
+#include "services.h"
+
 #include "msgq.hpp"
 
 void sigusr2_handler(int signal) {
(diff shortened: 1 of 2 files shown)


messaging/msgq.cc
```diff
@@ -81,10 +83,19 @@ void msgq_wait_for_subscriber(msgq_queue_t *q){
   return;
 }
 
-
+bool valid(std::string path){
```
pd0wm, line 86:
```suggestion
bool service_exists(std::string path){
```

messaging/msgq.cc
```diff
@@ -81,10 +83,19 @@ void msgq_wait_for_subscriber(msgq_queue_t *q){
...
+    if (it.name == path) {
+      return true;
+    }
+  }
+  return false;
+}
 
 int msgq_new_queue(msgq_queue_t * q, const char * path, size_t size){
   assert(size < 0xFFFFFFFF); // Buffer must be smaller than 2^32 bytes
+  if (!valid(std::string(path)))
```
pd0wm, line 97: put brackets

PR #88, opened 2020-09-14
Cross compilation arm64

sd-ricardo:
This PR is part of commaai/openpilot#2171 add is needed for a successful cross-compilation


SConstruct
```diff
@@ -39,6 +43,11 @@ env = Environment(
     "-O2",
     "-Wunused",
     "-Werror",
+    "-Werror=implicit-function-declaration",
```
adeebshihadeh, line 46: If you want to enable these, let's do it in a separate PR. also, this should match openpilot

adeebshihadeh:
Closing due to inactivity

PR #91, opened 2020-09-26
free reosurces after failed

deanlee:


pd0wm:
What is the benefit of deleting immediately over having the destructor clean up the `q`?

deanlee:
sorry it's my fault.

PR #92, opened 2020-10-03
Adding fields to check if the fan is spinning

LoganxDev:


pd0wm:
As mentioned in the openpilot PR. We only need the `fanNotSpinning` alert type.

adeebshihadeh:
Added the fanMalfunction event - 7755833e870e0fd80b1e70cbed195bfc0daff860

PR #93, opened 2020-10-04
Switch send to using bytes.

alexlitz:
Using string for send introduced a significant conversion overhead for large messages. For instance, for the below example based on demo.py I have observed a 2-4x speedup.
```

import time
import sys
import os
import cereal
from cereal import messaging
from cereal.messaging import Context, SubSocket, PubSocket, Poller
bts = ("a"*1280*720*3*4).encode("ascii")
MSGS = 1000

if __name__ == "__main__":
  c = Context()
  sub_sock = SubSocket()
  pub_sock = PubSocket()

  sub_sock.connect(c, "controlsState")
  pub_sock.connect(c, "controlsState")

  poller = Poller()
  poller.registerSocket(sub_sock)

  t = time.time()
  for i in range(int(MSGS)):
    pub_sock.send(bts)

    for s in poller.poll(100):
      dat = s.receive()

  dt = time.time() - t
  print("%.1f msg/s" % (MSGS / dt))
```

### DIFF
--- a/messaging/messaging_pyx.pyx
+++ b/messaging/messaging_pyx.pyx
@@ -140,9 +140,9 @@ cdef class PubSocket:
       else:
         raise MessagingError
 
-  def send(self, string data):
+  def send(self, bytes data):
     length = len(data)
-    r = self.socket.send(<char*>data.c_str(), length)
+    r = self.socket.send(<char*>data, length)
 
     if r != length:
       if errno.errno == errno.EADDRINUSE:
--- a/messaging/tests/test_poller.py
+++ b/messaging/tests/test_poller.py
@@ -97,16 +97,16 @@ class TestPoller(unittest.TestCase):
 
     time.sleep(0.1)  # Slow joiner
 
-    for i in range(100):
-      pub.send(str(i))
+    for i in range(1, 100):
+      pub.send(b'a'*i)
 
     msg_seen = False
-    i = 0
+    i = 1
     while True:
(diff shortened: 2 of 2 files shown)


pd0wm:
Good catch! I just fixed the tests on master to use bytes in the send call, we probably missed that in the python 2 -> python 3 migration. (https://github.com/commaai/cereal/commit/42e66ba92e1db918256b01b6b12f91e5d5f08dd0). If you rebase the tests will probably pass.

pd0wm:
If you have some time maybe you can do the same optimization in the can parser? https://github.com/commaai/opendbc/tree/master/can

alexlitz:
> Good catch! I just fixed the tests on master to use bytes in the send call, we probably missed that in the python 2 -> python 3 migration. ([42e66ba](https://github.com/commaai/cereal/commit/42e66ba92e1db918256b01b6b12f91e5d5f08dd0)). If you rebase the tests will probably pass.

Will do.

> If you have some time maybe you can do the same optimization in the can parser? https://github.com/commaai/opendbc/tree/master/can

I'll look into it.

Two things on separate notes, I am interested in [this](https://github.com/commaai/cereal/pull/78#issue-465219531) PR and I have a modified version of this repo with build modified to easily accommodate additional .capnp files and with demo.py, etc. moved to the tests folder (in a way such that they run properly). If it is of interest I can make a PR for those changes.

PR #98, opened 2020-11-08
Added kfV, kfBP for improved PIcontroller

nickn17:
Xps fork on HKG models used improved PIcontroler (kfV, kfBP instead kf). Steering on whole speed range and especially on lower speed up to 60kmh was much better that master or xx979x. I would like to use it for Kia Niro EV.

https://github.com/commaai/openpilot/discussions/2401#discussioncomment-123170

### DIFF
--- a/car.capnp
+++ b/car.capnp
@@ -421,6 +421,8 @@ struct CarParams {
     kiBP @2 :List(Float32);
     kiV @3 :List(Float32);
     kf @4 :Float32;
+    kfV @5 :List(Float32);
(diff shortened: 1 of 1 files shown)


gnarf0815:
I absolutely agree. Would love to have that improvement on stock openpilot. I think it would be a great enhancement.
Thank you all for your efforts.

adeebshihadeh:
We're looking into improving controls in a way that would make manual tuning obsolete. In the meantime, if you want more parameters to play with, try INDI.

PR #101, opened 2020-11-16
Visionipc v2.0

pd0wm:


.gitignore
```diff
@@ -6,7 +6,12 @@ package-lock.json
 __pycache__
 .*.swp
 .*.swo
+*.os
+*.o
+
+visionipc/test
 libcereal*.a
+libvisionipc.a
```
adeebshihadeh, line 14: should probably just ignore all *.a static libraries, don't think we have any checked in here

visionipc/visionipc.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+#include <unistd.h>
+#include <stdint.h>
+
+#define VISIONIPC_MAX_FDS 64
+
+#ifdef __cplusplus
```
adeebshihadeh, line 8: do we need this anymore? everything that uses vipc should be c++ already

visionipc/visionbuf_cl.cc
```diff
@@ -0,0 +1,93 @@
...
+#include <unistd.h>
+#include <sys/mman.h>
+#include <sys/types.h>
+
+#define CL_USE_DEPRECATED_OPENCL_1_2_APIS
+#ifdef __APPLE__
+#include <OpenCL/cl.h>
+#else
+#include <CL/cl.h>
+#endif
```
adeebshihadeh, line 16: already defined in visionbuf.h

visionipc/ipc.cc
```diff
@@ -0,0 +1,121 @@
...
+#include <sys/mman.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+
+#include "ipc.h"
+
+int ipc_connect(const char* socket_path) {
+  int err;
+
+#ifdef __APPLE__
```
adeebshihadeh, line 17: do this ifdef once on top instead of repeating

visionipc/visionbuf_cl.cc
```diff
@@ -0,0 +1,93 @@
...
+
+int offset = 0;
```
adeebshihadeh, line 18: this global isn't great, can easily see shadowing bugs from this

adeebshihadeh: also, make visionbuf a cpp class?

pd0wm: The global was already there in old visionipc. If I turn it into an std::atomic<int> I think it should be safe. If you fork the process you get a different pid, so that should also be ok.

visionipc/visionipc_server.h
```diff
@@ -0,0 +1,42 @@
...
+
+  Context * msg_ctx;
+  std::map<VisionStreamType, PubSocket*> sockets;
+
+  void listener(void);
+  void init(void);
+
+ public:
+  VisionIpcServer(std::string name, cl_device_id device_id, cl_context ctx);
+  VisionIpcServer(std::string name);
```
adeebshihadeh, line 34: do we need two constructors? can't we just make device_id and ctx default to nullptr? then we can also get rid of the init function

visionipc/visionipc.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include <cstdint>
+#include <cstddef>
+
+#define VISIONIPC_MAX_FDS 64
+
+struct VIPCBufExtra {
```
adeebshihadeh, line 8: VIPC -> VisionIpc for consistency

visionipc/visionipc.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include <cstdint>
+#include <cstddef>
+
+#define VISIONIPC_MAX_FDS 64
```
adeebshihadeh, line 6:
```suggestion
constexpr int VISIONIPC_MAX_FDS = 64;
```

visionipc/visionbuf_ion.cc
```diff
@@ -0,0 +1,151 @@
...
+  switch (dir) {
+  case VISIONBUF_SYNC_FROM_DEVICE:
+    custom_data.cmd = ION_IOC_INV_CACHES;
+    break;
+  case VISIONBUF_SYNC_TO_DEVICE:
+    custom_data.cmd = ION_IOC_CLEAN_CACHES;
+    break;
+  default:
+    assert(0);
+  }
```
adeebshihadeh, line 131:
```suggestion
   assert(dir == VISIONBUF_SYNC_FROM_DEVICE || dir == VISIONBUF_SYNC_TO_DEVICE);
   custom_data.cmd = dir == VISIONBUF_SYNC_FROM_DEVICE ?
                                      ION_IOC_INV_CACHES : ION_IOC_CLEAN_CACHES;
```

visionipc/visionipc_server.cc
```diff
@@ -0,0 +1,175 @@
...
+VisionIpcServer::VisionIpcServer(std::string name) : name(name) {
+  init();
+}
+
+void VisionIpcServer::init(void){
+  msg_ctx = Context::create();
+
+  std::random_device rd("/dev/urandom");
+  std::uniform_int_distribution<uint64_t> distribution(0,std::numeric_limits<uint64_t>::max());
+  server_id = distribution(rd);
```
adeebshihadeh, line 28: can't we use the PID or do we want to support multiple servers in a process?

pd0wm: I'd like to support multiple servers in one process

PR #104, opened 2020-12-05
Added stoppingBrakeRate to CarParams

briskspirit:
Added stoppingBrakeRate to CarParams to parametrize STOPPING_BRAKE_RATE to car interfaces in OpenPilot.


pd0wm:
I'll merge this once we're sure we actually need to change this specifically for TSS2 Toyota.

briskspirit:
Few people and I tried BRAKE_STOPPING_RATE = 0.1 and everybody agree that it is much smoother on stop, this can't be solved by long tune itself. But this const is used by all cars and I can't be sure that it will be the same for every car(but I assume so). So there is 2 ways, right? To change this value for every car or for now to change it only for TSS2 cars by parametrizing it.

PR #105, opened 2020-12-10
add warning when car is about to disengage adaptive cruise

cydia2020:
So that user can take appropriate actions


adeebshihadeh:
Closing since the matching openpilot PR was closed.

PR #107, opened 2021-01-03
Best practice

HaraldSchafer:


HaraldSchafer:
Log name best practice violations
- [logMonoTime]
  Not SI units, value not correctly described. Could be bootTimeNanos or nanosSinceBoot
  
 - [Frame, FrontFrame, WideFrame]
 Totally unclear. Could be RoadFrame,  RoadFrameWide, DriverFrame ?
 
 - [GpsLocationData.bearing, GpsLocationData.bearingAccuracy]
 Not SI units. Could be bearingDegrees

- [ControlsState.angleSteers, ControlsState.angleSteersDes, LateralINDIState.steer, LateralPIDState.steer, LateralLQRState]
Is this in degrees? Not SI units. What is angleSteer anyway? Angle of tires or steering wheel?

- [PathPlan.angleOffset, LiveParametersData.angleOffset]
Is this in degrees? Not SI units.

service_list.yaml
```diff
@@ -81,6 +81,17 @@ wideEncodeIdx: [8075, true, 20.]
 wideFrame: [8076, true, 20.]
 modelV2: [8077, true, 20., 20]
 
+sensorAccelerometer1: [8078, true, 20, 1]
+sensorAccelerometer2: [8079 , true, 20, 1]
```
adeebshihadeh, line 85: these frequencies are wrong. do we want them in qlogs?

pd0wm: This was just a proposal. Need to look up the frequencies and we need to decide what we want to keep for each sensor.

adeebshihadeh: think we want to do this in another PR anyway

HaraldSchafer:
**ANDROIDSENSOR**
- [ ] What is this? What kind of sensors can this be?
**ThermalData**
- [ ] This isn't thermal stuff?
- [ ] freeSpace? Units? freeSpacePercent?
- [ ] battery? Is this deviceBattery?
- [ ] offroad? standbyPowerUsage better?
- [ ] carBatteryCapicty, not SI units,put unit in name
- [ ] fanSpeed? units? Make fanRPM?
- [ ] startedTs? What is this? Make startTimeNanos?
- [ ] cpuPerc? Make cpuPercent so like memUsedPercent?
- [ ] thermalStatus? This has nothing to do with thermal
- [ ] cpu, gpu, mem, bat, add units and temp

HaraldSchafer:
**HealthData**

- [ ] voltage and current of what? Is all of this stuff about the car? If so, it is not clear. Also hwType does not refer to car. Could be carInterfaceHw or something?
- [ ] Deprecated should be capitalized

**RadarState**
- [ ]  This message will be deprecated with the longi rework

**LiveCalibrationData**
- [ ] Matrices should be deprecated

mitchellgoffpc:
Struct fields in car.capnp that I didn't immediately understand:

mitchellgoffpc:
Complaints for log.capnp:
- ThermalData: Ditto what harald said
- ControlsState: What is angleSteersDes?
- ControlsState.OpenpilotState: Curious what preEnabled means
- Seems like Plan and PathPlan should be renamed to LongitudinalPlan and LateralPlan, respectively

ZwX1616:
- Why do we have both DeviceType in InitData and HwType in HealthData? And the types are different?
- All timestamps: unit is unclear, may just add a comment saying they are nanos if not specified
- FrameData: integLines is murky and hardly useful, should add a field like expTimeMillis; FrameType - neo/chffr/front?? should just deprecate
- SensorEventData: every sensor should be documented like GpsLocationData
- ditto ThermalData cleanup
- ControlsState: needs cleanup similar to thermal? at least move relevant ones into structs?
- Model: at least explain xyz for easy parsing
- ProcLog: cpuUser -> cpuTimeUser? etc.
- DriverState/sgProb: unclear, will fix

adeebshihadeh:
@deanlee you're pretty familiar with our structs, any additional complaints?

deanlee:
Suggestions:
- [ ] Moving the alert to a separate structure

legacy.capnp
```diff
@@ -200,6 +200,72 @@ struct UiNavigationEvent @0x90c8426c3eaddd3b {
...
+  gpsWeek @16 :Int32;
+  timeOfWeek @17 :Float64;
+
+  positionECEF @18 :List(Float64);
+  poseQuatECEF @19 :List(Float32);
+  pitchCalibration @20 :Float32;
+  yawCalibration @21 :Float32;
+  imuFrame @22 :List(Float32);
+
+  struct Accuracy {
```
adeebshihadeh, line 249: struct and enums need an explicit id as well

messaging/__init__.py
```diff
@@ -14,7 +14,7 @@
 # sec_since_boot is faster, but allow to run standalone too
 try:
   from common.realtime import sec_since_boot
-except ImportError:
+except Exception:
```
adeebshihadeh, line 17: was this an issue? ideally, we only watch to catch ImportError

log.capnp
```diff
@@ -296,12 +296,12 @@ struct ThermalData {
   chargingError @17 :Bool;
   chargingDisabled @18 :Bool;
 
-  # temperatures
-  cpu @26 :List(Float32);
-  gpu @27 :List(Float32);
-  mem @28 :Float32;
-  bat @29 :Float32;
-  ambient @30 :Float32;
+  # temperatures TODO units?
```
adeebshihadeh, line 299: these are all celsius

legacy.capnp
HaraldSchafer: ok

messaging/__init__.py
HaraldSchafer: Traceback (most recent call last):                                                                                                       
  File "./test_messaging.py", line 13, in <module>                                                                                       
    import cereal.messaging as messaging                                                                                                 
  File "/home/batman/openpilot/cereal/messaging/__init__.py", line 16, in <module>                
    from common.realtime import sec_since_boot                
  File "/home/batman/openpilot/common/realtime.py", line 8, in <module>                     
    from selfdrive.hardware import PC, TICI                                                                                              
  File "/home/batman/openpilot/selfdrive/hardware/__init__.py", line 5, in <module>   
    from selfdrive.hardware.eon.hardware import Android                                                                                  
  File "/home/batman/openpilot/selfdrive/hardware/eon/hardware.py", line 11, in <module>             
    NetworkType = log.ThermalData.NetworkType                                                                                            
AttributeError: module 'log.capnp' has no attribute 'ThermalData' 


was getting that

log.capnp
HaraldSchafer: ok

messaging/__init__.py
adeebshihadeh: that's because `selfdrive/hardware/eon/hardware.py` needs to be updated to the DeviceState struct

adeebshihadeh: we don't want to make this catch all exceptions, otherwise we'll never know if we break this unintentionally and always end up using the slower function

HaraldSchafer: I know, but we can't do that without merging this PR into OP first?

legacy.capnp
HaraldSchafer: If this is needed, can we make the tests fail on this?

messaging/__init__.py
adeebshihadeh: the CI tests don't rely on openpilot, so those should be fine. if you check out a matching cereal cleanup openpilot branch, the tests should pass locally too

HaraldSchafer: Oh I thought this was broken in CI too, ok will revert

legacy.capnp
adeebshihadeh: yeah, i can write a backwards compatibility test before we merge this

log.capnp
```diff
@@ -296,12 +296,11 @@ struct ThermalData {
   chargingError @17 :Bool;
   chargingDisabled @18 :Bool;
 
-  # temperatures
-  cpu @26 :List(Float32);
-  gpu @27 :List(Float32);
-  mem @28 :Float32;
-  bat @29 :Float32;
-  ambient @30 :Float32;
+  cpuCelsius @26 :List(Float32);
```
adeebshihadeh, line 299: maybe xxxTempC?

adeebshihadeh:
- [ ] androidLog - this can now be either systemd or logcat

log.capnp
HaraldSchafer: ok

log.capnp
```diff
@@ -274,16 +274,16 @@ struct CanData {
   src     @3 :UInt8;
 }
 
-struct ThermalData {
+struct DeviceStateData {
```
adeebshihadeh, line 277: need an explicit id

log.capnp
```diff
@@ -339,7 +338,7 @@ struct ThermalData {
   pa0DEPRECATED @21 :UInt16;
 }
 
-struct HealthData {
+struct PandaStateData {
```
adeebshihadeh, line 341: need an explicit id

PR #108, opened 2021-01-05
Adding breakpoints to INDI lateral tuning

briskspirit:
Adding breakpoints to INDI lateral tuning


pd0wm:
Can you keep the old ones but add `DEPRECATED` as a suffix? Otherwise it makes old logs unreadable.

briskspirit:
Done, thank you.

PR #112, opened 2021-01-23
SubMaste::updated :  return std::optional<cereal::Event::Reader>

deanlee:


adeebshihadeh:
Left a comment on the openpilot PR.

PR #113, opened 2021-01-24
split out deprecated structs into separate schema file

adeebshihadeh:


log.capnp
```diff
@@ -2095,56 +2039,67 @@ struct Event {
...
+    liveUIDEPRECATED @14 :Legacy.LiveUI;
+    sensorEventDEPRECATED @4 :SensorEventData;
+    liveEventDEPRECATED @8 :List(Legacy.LiveEventData);
+
+    ethernetData @26 :List(Legacy.EthernetPacket);
+    cellInfo @28 :List(Legacy.CellInfo);
+    wifiScan @29 :List(Legacy.WifiScan);
+
+    # legacy?
+    liveMapData @62 :LiveMapData;
```
adeebshihadeh, line 2085: @HaraldSchafer these all good to deprecate?

HaraldSchafer: yes, don't know anything about ethernetData, cellinfo and wifiScan  but others are good to go

HaraldSchafer:
Anything blocking this merge?

adeebshihadeh:
This isn't backwards compatible. capnproto automatically generates an id for each struct, and we have to copy that implicit id and make it explicit like we do with controlsState.

PR #114, opened 2021-01-26
handle signals gracefully

deanlee:


pd0wm:
Support for blocking receive seems to bring with it this big mess of signaling to the caller that the operation was interrupted by a SIGINT. I'm thinking of removing support for this altogether and only support pollers and non blocking receive. Are there places where we absolutely need a blocking receive?

deanlee:
I'll second that,  It's best not to have blocking mode.

PR #125, opened 2021-03-14
Added new util class AlignedBuffer

deanlee:


messaging/messaging.hpp
```diff
@@ -120,3 +120,20 @@ class PubMaster {
 private:
   std::map<std::string, PubSocket *> sockets_;
 };
+class AlignedBuffer {
+public:
+  AlignedBuffer() = default;
+  AlignedBuffer(const char *data, const size_t size) { get(data, size); }
+  kj::ArrayPtr<const capnp::word> get(const char *data, const size_t size) {
```
pd0wm, line 127: Maybe add another overloaded method that takes in a `Message`?

pd0wm: e.g. `kj::ArrayPtr<const capnp::word> get(const Message msg) { return get(msg->getData(), msg->getSize()); }`

messaging/messaging.hpp
```diff
@@ -120,3 +120,20 @@ class PubMaster {
 private:
   std::map<std::string, PubSocket *> sockets_;
 };
+class AlignedBuffer {
```
pd0wm, line 123: whitespace

PR #131, opened 2021-04-08
MessageBuilder: use stack array for the first segment

deanlee:
related document from capnp:
```
explicit MallocMessageBuilder(kj::ArrayPtr<word> firstSegment,   AllocationStrategy allocationStrategy = SUGGESTED_ALLOCATION_STRATEGY);
  // This version always returns the given array for the first segment, and then proceeds with the
  // allocation strategy.  This is useful for optimization when building lots of small messages in
  // a tight loop:  you can reuse the space for the first segment.
  //
  // firstSegment MUST be zero-initialized.  MallocMessageBuilder's destructor will write new zeros
  // over any space that was used so that it can be reused.

```

### DIFF
--- a/messaging/messaging.hpp
+++ b/messaging/messaging.hpp
@@ -87,9 +87,13 @@ private:
   std::map<std::string, SubMessage *> services_;
 };
 
-class MessageBuilder : public capnp::MallocMessageBuilder {
+struct MessageBuilderSegmentBuffer {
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Still relevant or can we close?

deanlee:
This implementation is not good. closed.

PR #133, opened 2021-04-08
SubMaster: split socket recv and update

iejMac:


messaging/socketmaster.cc
```diff
@@ -66,33 +69,50 @@ int SubMaster::update(int timeout) {
...
-    m->rcv_time = current_time;
-    m->rcv_frame = frame;
-    m->valid = m->event.getValid();
 
-    ++updated;
+    messages.push_back(m->msg_reader->getRoot<cereal::Event>());
+  }
+
+  update_msgs(current_time, messages);
+  return messages.size();
```
adeebshihadeh, line 93: let's match the python API and return nothing here

messaging/socketmaster.cc
```diff
@@ -62,37 +64,47 @@ SubMaster::SubMaster(const std::initializer_list<const char *> &service_list, co
   }
 }
 
-int SubMaster::update(int timeout) {
+void SubMaster::update(int timeout) {
   if (++frame == UINT64_MAX) frame = 1;
```
adeebshihadeh, line 68: move this to update_msgs

messaging/messaging.hpp
```diff
@@ -67,7 +67,8 @@ class SubMaster {
 public:
   SubMaster(const std::initializer_list<const char *> &service_list,
             const char *address = nullptr, const std::initializer_list<const char *> &ignore_alive = {});
-  int update(int timeout = 1000);
+  void update(int timeout = 1000);
+  void update_msgs(int current_time, std::map<std::string, cereal::Event::Reader> messages);
```
adeebshihadeh, line 71:
```suggestion
  void update_msgs(uint64_t current_time, std::map<std::string, cereal::Event::Reader> messages);
```

PR #136, opened 2021-04-19
Automatically generate service ports

ShaneSmiskol:
These might be out of date, I heard you're switching to 22 for ssh? Let me know if these are incorrect and I can update/close.

This would help me and probably others from adding a new struct here that accidentally overlapped a reserved port

### DIFF
--- a/messaging/tests/test_services.py
+++ b/messaging/tests/test_services.py
@@ -5,7 +5,7 @@ import unittest
 from parameterized import parameterized
 
 import cereal.services as services
-from cereal.services import service_list
+from cereal.services import service_list, RESERVED_PORTS
 
 
 class TestServices(unittest.TestCase):
@@ -14,6 +14,7 @@ class TestServices(unittest.TestCase):
   def test_services(self, s):
     service = service_list[s]
     self.assertTrue(service.port > 8000)
+    self.assertTrue(service.port not in RESERVED_PORTS)
     self.assertTrue(service.frequency <= 100)
 
   def test_no_duplicate_port(self):
--- a/messaging/tests/test_services.py
+++ b/messaging/tests/test_services.py
@@ -5,7 +5,7 @@ import unittest
 from parameterized import parameterized
 
 import cereal.services as services
-from cereal.services import service_list, RESERVED_PORTS
+from cereal.services import service_list, RESERVED_PORTS, STARTING_PORT, MAX_FREQ
 
 
 class TestServices(unittest.TestCase):
@@ -13,9 +13,9 @@ class TestServices(unittest.TestCase):
   @parameterized.expand(service_list.keys())
   def test_services(self, s):
     service = service_list[s]
-    self.assertTrue(service.port > 8000)
+    self.assertTrue(service.port > STARTING_PORT - 1)
     self.assertTrue(service.port not in RESERVED_PORTS)
-    self.assertTrue(service.frequency <= 100)
+    self.assertTrue(service.frequency <= MAX_FREQ)
 
   def test_no_duplicate_port(self):
     ports = {}
--- a/messaging/tests/test_services.py
+++ b/messaging/tests/test_services.py
@@ -5,7 +5,7 @@ import unittest
 from parameterized import parameterized
 
 import cereal.services as services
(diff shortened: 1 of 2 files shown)


pd0wm:
8080 and 8762 are not used anymore, so those checks can be removed. Think we'll be running SSH on both 8022 and 22 for a while, so let's leave that check.

ShaneSmiskol:
Done!
I don't know if this would make sense, but do the ports need to be manually defined and static? We could just have a counter that starts at 8000 and skips the reserved ports for each new service, assuming we add to the bottom it wouldn't change the above service ports adding a new one. Might be unnecessary?

pd0wm:
Yeah let's automatically generate them! We only use them for zmq anyway, so not that important.

services.py
```diff
@@ -4,61 +4,75 @@
 
 EON = os.path.isfile('/EON')
 
+RESERVED_PORTS = [8022, 22]  # sshd
+STARTING_PORT = 8001
```
ShaneSmiskol, line 8: test_services.py checked that port > 8000, can we do >= 8000 instead or is that reserved?

services.py
```diff
@@ -4,61 +4,75 @@
...
+
+
+cur_port = STARTING_PORT
+def new_port():
```
pd0wm, line 12: Don't like the global. Why not keep the port argument to `Service`, but build the dict dynamically.

E.g. something like
```python
services = {
  "roadCameraState": (True, 20., 1),
  "sensorEvents": (True, 100., 100),
}
service_list = {name: Service(i + 8000, *vals) for i, (name, vals) in enumerate(sevices.items())}
```

services.py
pd0wm: Not exactly sure. Let's keep 8001 as a starting port.

services.py
ShaneSmiskol: I couldn't figure out how to create a function that could skip the reserved ports while providing continuous ports just with the index without knowing the previous port (or back calculating it starting from 8001, which takes exponentially more time as we add services) so I relied on the handy walrus operator to do this.

ShaneSmiskol: I guess since you don't care what the ports are we could just add an offset to the ports that happen to be reserved, something like the length of the `services` dict so that it doesn't collide with the last service port.

services.py
```diff
@@ -4,62 +4,73 @@
 
 EON = os.path.isfile('/EON')
 
+RESERVED_PORTS = {8022, 22}  # sshd
+STARTING_PORT = 8001
+
+
+def new_port(port: int):
+  port += STARTING_PORT
+  return port + len(services) if port in RESERVED_PORTS else port
```
pd0wm, line 13: Let's not make it too complicated. We only need to reserve 8022. So this can just be
```python
RESERVED_PORT = 8022
STARTING_PORT = 8001

def new_port(idx: int):
  port = STARTING_PORT + idx
  return port + 1 if port >= RESERVED_PORT else port
```

ShaneSmiskol: Okay gotcha. I just wanted to make sure it would be easily extensible if we needed many

pd0wm: If the general solution is more complex/hard to read it's not really worth it. If we need more ports we can always change the implementation later.

ShaneSmiskol: I think there's a bug with mypy, I saw a few similar issues with unpacking a tuple for a function. I'll try adding a comment to ignore the line

messaging/tests/test_services.py
```diff
@@ -13,7 +13,6 @@ class TestServices(unittest.TestCase):
   @parameterized.expand(service_list.keys())
   def test_services(self, s):
     service = service_list[s]
-    self.assertTrue(service.port > 8000)
```
pd0wm, line 16: Why not leave this check? Import STARTING_PORT and compare with that as a sanity check.

pd0wm: Also add a check the reserved port is not used.

ShaneSmiskol: Right, reverted! I think I was brainwashed to optimize for lines haha, sorry. Is there anything else for this PR?

PR #152, opened 2021-05-19
Add subaruGen2 and subaruHybrid to CarParams

martinl:
- subaruGen2 is for upcoming Outback/Legacy 2020 port
- subaruHybrid is for upcoming Crosstrek Hybrid 2020 port

Both have slightly different CAN network topology vs Global Platform - some of the messages required by openpilot are new and/or on can1. I will open panda PR for review soon

### DIFF
--- a/car.capnp
+++ b/car.capnp
@@ -501,6 +501,8 @@ struct CarParams {
     subaruLegacy @22;  # pre-Global platform
     hyundaiLegacy @23;
     hyundaiCommunity @24;
+    subaruGen2 @25;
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Do we have car ports for these yet? I don't want to add them until a car port is ready.

martinl:
Both are still work in progress, full openpilot support will require external panda and custom car harness to intercept both can0 and can1 at camera connector

adeebshihadeh:
Let's revisit this once we have car port PRs.

PR #160, opened 2021-05-29
Unbridge - for republishing msgs sent from your laptop to openpilot

ShaneSmiskol:
Unbridge - for republishing msgs sent from your laptop to openpilot. Just start bridge via: `./bridge {LAPTOP_IP}`

Todo:
- [x] Combine bridge and unbridge with an argument
- [x] Even if you don't send anything, creating another publisher causes a `MultiplePublishersError`, so:
  - [x] ~~See about passing a service list, OR...~~ can do this, but for now doing the latter: (also thought about creating all sub sockets and when we receive a message create a pub socket dynamically, but it was too complex)
  - [x] ...when passed `--reverse` only create sockets for the services with the `test` prefix, and ignore everything else'
- [x] look into `getopt` for arg parsing

Goes with: https://github.com/commaai/openpilot/pull/21129


messaging/bridge.cc
```diff
@@ -29,26 +29,41 @@ static std::vector<std::string> get_services() {
 }
 
 
-int main(void){
+int main(int argc, char** argv){
   signal(SIGPIPE, (sighandler_t)sigpipe_handler);
 
+  bool unbridge = argc > 1 && strcmp(argv[1], "--unbridge") == 0;
```
ShaneSmiskol, line 35: --reverse?

ClockeNessMnstr:
Joystickd param call fails params_pyx can't find testJoystick or something

Events are causing LKAS faults and not showing up. 

I manually copied over the changes into master and could have messed it up. 

Got me going though.

PR #172, opened 2021-06-27
submaster: revert changes in update

deanlee:
revert  `submaster::update` to the origin version.
I think it's better not to change  `SubMaster::update` for `update_msgs`. it increased complexity and reduced performance.


adeebshihadeh:
It was split to support replay without sockets.

deanlee:
If i didn't understand it wrong,these changes should not affect replay without sockets.the `update_msgs` is still the same as before.

PR #188, opened 2021-08-15
SubMaster: set traversalLimitInWords to max

deanlee:


adeebshihadeh:
Let's do this for the python SubMaster too.

messaging/__init__.py
```diff
@@ -83,7 +84,7 @@ def drain_sock(sock: SubSocket, wait_for_one: bool = False) -> List[capnp.lib.ca
     if dat is None:  # Timeout hit
       break
 
-    dat = log.Event.from_bytes(dat)
+    dat = log.Event.from_bytes(dat, traversal_limit_in_words=NO_TRAVERSAL_LIMIT)
```
pd0wm, line 87: Can you make a helper function `log_from_bytes(dat)`?

PR #191, opened 2021-08-28
VisionBuf: retry ioctls while errno == EINTR

deanlee:
relating to https://github.com/commaai/openpilot/pull/22054


adeebshihadeh:
LGTM. any comments @pd0wm?

PR #194, opened 2021-09-11
Stellantis: Allocate safety mode ID

jyoung8607:
Prep for upstream of Tunder's 5th gen RAM 1500 port, which may also support the new 4th gen Jeep Wagoneer and Grand Wagoneer. Messaging and operation looks to be fully distinct from the older Chrysler/FCA platforms.


adeebshihadeh:
Looking forward to the openpilot PR!

PR #196, opened 2021-09-12
split pandaState into pandaState and peripheralState

pd0wm:


log.capnp
```diff
@@ -454,6 +447,21 @@ struct PandaState @0xa7649e2575e4591e {
   startedSignalDetectedDEPRECATED @5 :Bool;
 }
 
+struct PeripheralState {
+  voltage @0 :UInt32;
+  current @1 :UInt32;
+  fanSpeedRpm @2 :UInt16;
+  powerSaveEnabled @3 :Bool;
```
adeebshihadeh, line 454: shouldn't this stay per-panda?

robbederks:
Not really sure what you're trying to accomplish by splitting these out like this. Wouldn't it suffice to just do something like this:
```
masterPandaState :PandaState;
peripheralPandaStates :List(PandaState);
```
or even just 

```
pandaStates :List(PandaState);
```
where the first entry will always be the internal one?

pd0wm:
The goal is to split driving related tasks from what we hacked into the panda state to control the device. Think this is a good change regardless of multiple panda support.

It also makes the upcoming changes to boardd easier to reason about since you can have two different threads and you can pass the appropriate panda to each one.

log.capnp
```diff
@@ -393,6 +387,12 @@ struct PandaState @0xa7649e2575e4591e {
   harnessStatus @21 :HarnessStatus;
   heartbeatLost @22 :Bool;
 
+  voltageDEPRECATED @0 :UInt32;
```
adeebshihadeh, line 390: deprecated fields usually go at the bottom

PR #202, opened 2021-09-27
remove androidLog from qlogs

adeebshihadeh:
@pd0wm last thing that relies on this is "LTE emm-detached" detection. do we still need that? maybe we should make it a cloudlog instead?


pd0wm:
cloudlog should be fine for that.

PR #204, opened 2021-10-05
Add STD to the liveParameter message to enable more accurate offline runs

nuwandavek:
Currently when we run offline `parmasd`, the covariances are always initialized to incorrect default values. Due to this, we cannot currently recreate online failures accurately. 

I did not create a new `Measurement` struct like in `liveLocationKalman` as some of the variables are expected in degrees and I wanted to avoid needless conversions.

### DIFF
--- a/log.capnp
+++ b/log.capnp
@@ -1300,10 +1300,10 @@ struct LiveParametersData {
   yawRate @7 :Float32;
   posenetSpeed @8 :Float32;
   posenetValid @9 :Bool;
-  angleOffsetFastSTD @10 :Float32;
-  angleOffsetAverageSTD @11 :Float32;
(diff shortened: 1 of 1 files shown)


log.capnp
```diff
@@ -1300,6 +1300,10 @@ struct LiveParametersData {
   yawRate @7 :Float32;
   posenetSpeed @8 :Float32;
   posenetValid @9 :Bool;
+  angleOffsetFastSTD @10 :Float32;
```
adeebshihadeh, line 1303: this is the convention used elsewhere in cereal
```suggestion
  angleOffsetFastStd @10 :Float32;
```

adeebshihadeh:
Yes, the failing test has been fixed on master.

PR #214, opened 2021-10-31
fix wshadow errors

mayfieldiv:
See https://github.com/commaai/openpilot/pull/22756

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -45,6 +45,7 @@ env = Environment(
     "-O2",
     "-Wunused",
     "-Werror",
+    "-Wshadow",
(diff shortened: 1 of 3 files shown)


pd0wm:
Can you add the flag to the `cereal` sconstruct so we catch any errors immediately? Same for `rednose` and `opendbc`.

mayfieldiv:
@pd0wm, sorry you'll need to allow the gitlab action to run again

pd0wm:
Updated this repo's action settings to match our other repos. So should be approved now by default since you're not a new github user.

PR #225, opened 2021-11-26
add carParams.flags

adeebshihadeh:


belm0:
👏   have been looking for something like this

car.capnp
```diff
@@ -379,6 +379,7 @@ struct CarParams {
   enableDsu @5 :Bool;        # driving support unit
   enableApgs @6 :Bool;       # advanced parking guidance system
   enableBsm @56 :Bool;       # blind spot monitoring
+  flags @63 :UInt32;         # flags for car specific quirks
```
belm0, line 382: I'm trying to think of a name that conveys this is car-specific.  (And by "car-specific quirks", we mean the flag definitions are local to a subset of cars, such as a folder under openpilot `selfdrive/car`.)

How about `localFlags`?

belm0: and is it possible other values may be packed into these bits?  (integer magnitude, etc.)  Consider dropping "flags":

```python
custom @63 :UInt32;        # custom bits local to a subset of cars
custom2 @64 :UInt32;
```

PR #233, opened 2021-12-15
Log actual actuator command that was sent

pd0wm:


car.capnp
```diff
@@ -291,6 +291,7 @@ struct CarControl {
   active @7 :Bool;
 
   actuators @6 :Actuators;
+  actuatorsFinal @10 :Actuators;
```
pd0wm, line 294: @adeebshihadeh Any suggestions for a better name?

adeebshihadeh: maybe actuatorsOutput? I'd also add comments to explain both

PR #234, opened 2021-12-19
Add hondaRadarless safety mode

vanillagorillaa:
2022 Civic will require a new panda safety mode and thus needs new enum


adeebshihadeh:
We'll likely just end up making this a safety param instead.

PR #237, opened 2022-01-04
AlignedBuffer: falls back to a copy if the buffer is not aligned

deanlee:
no copy if the buffer is already aligned (just like what pycapnp did)


adeebshihadeh:
Still relevant or can we close?

deanlee:
closed

PR #240, opened 2022-01-08
submaster: optimize pycapnp reader

deanlee:
cache data on-demand for faster access 

The performance of the first reading is the same as  pyapnp,  the subsequent reading is native performance and dozens faster than pycapnp, 

I believe this PR can greatly improve the performance of repeated data reading.

### DIFF
--- a/messaging/__init__.py
+++ b/messaging/__init__.py
@@ -4,7 +4,7 @@ from .messaging_pyx import MultiplePublishersError, MessagingError  # pylint: di
 import os
 import capnp
 
-from typing import Optional, List, Union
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
This looks like a good idea, but I'd like to make sure we're actually getting what we expect. Can you profile the following scenarios with a couple different structs and scenarios?
* read from namedtuple (as the benchmark)
* read from SubMaster without this change
* read from SubMaster with this change

Maybe use controlsState or carState and make sure to check each type: int, list, nested struct, etc.

PR #242, opened 2022-01-17
Add unsafeMode to carParams

sshane:
For https://github.com/commaai/openpilot/pull/23538 🙂

### DIFF
--- a/car.capnp
+++ b/car.capnp
@@ -398,6 +398,7 @@ struct CarParams {
   minSteerSpeed @8 :Float32;
   maxSteeringAngleDeg @54 :Float32;
   safetyConfigs @62 :List(SafetyConfig);
+  unsafeMode @65 :Int16;
 
   steerMaxBP @11 :List(Float32);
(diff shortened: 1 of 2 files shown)


car.capnp
```diff
@@ -458,6 +458,7 @@ struct CarParams {
   struct SafetyConfig {
     safetyModel @0 :SafetyModel;
     safetyParam @1 :Int16;
+    unsafeMode @2 :Int16;
```
adeebshihadeh, line 461: shouldn't go in this struct as it doesn't vary per-panda

sshane: Ah I see. It should be added to pandaState as well so we can verify what's set on the panda side as well, right?

sshane: On the panda side, we set each panda to have the unsafe_mode. So are we assuming that the function (`set_unsafe_mode` in panda.cc) will never fail? If it does, how do we know which panda to compare its unsafe_mode with the single one in carParams?

PR #243, opened 2022-01-22
VisionIpcClient/connect: return false if an invalid stream type is requested

deanlee:
The `VisionIpcClient::connect` should return false if an invalid stream type is requested. the server has closed the socket in this case, and the next call to ipc_sendrecv_with_fds will return 0.

This issue may cause the client to crash when a camera is not started. for example, start the `reply --demo`  without dcamera, then start the loggerd and it will crash.


adeebshihadeh:
Still relevant?

deanlee:
It looks like this is no longer an issue now.

PR #245, opened 2022-02-15
Add enum for car make

sshane:
Also effectively renames `carFingerprint`to `carModel` as an int, describing the new CAR enums in openpilot

### DIFF
--- a/car.capnp
+++ b/car.capnp
@@ -383,8 +383,8 @@ struct CarControl {
 # ****** car param ******
 
 struct CarParams {
-  carName @0 :Text;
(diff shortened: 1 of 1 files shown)


car.capnp
```diff
@@ -383,8 +383,8 @@ struct CarControl {
 # ****** car param ******
 
 struct CarParams {
-  carName @0 :Text;
-  carFingerprint @1 :Text;
+  carMake @50 :CarMake;
+  carModel @1 :UInt16;
```
sshane, line 387: Keep carFingerprint?

PR #262, opened 2022-03-21
Add unused attribute to fix scons gcc compile errors

JasonJShuler:
scons -u was failing due to several variables assigned but unused:

image

Applying the `__attribute__((unused))` attribute to these three instances resolves the build errors and allows the repo to compile:

image

### DIFF
--- a/messaging/msgq.cc
+++ b/messaging/msgq.cc
@@ -318,10 +318,10 @@ int msgq_msg_ready(msgq_queue_t * q){
     goto start;
   }
 
-  uint32_t read_cycles, read_pointer;
+  uint32_t __attribute__((unused))read_cycles, read_pointer;
   UNPACK64(read_cycles, read_pointer, *q->read_pointers[id]);
 
-  uint32_t write_cycles, write_pointer;
+  uint32_t __attribute__((unused))write_cycles, write_pointer;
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
This seems to just patch over a false positive from your compiler, possibly due to those variables only being passed to a macro.

You've made a few PRs like this fixing some real issues, but I'd recommend sticking to Ubuntu 20.04, our primary development target, for the best development experience. In this case, I'm not going to just merge this "zero impact" change just to fix platforms we don't target.

JasonJShuler:
> This seems to just patch over a false positive from your compiler, possibly due to those variables only being passed to a macro.
> 
> You've made a few PRs like this fixing some real issues, but I'd recommend sticking to Ubuntu 20.04, our primary development target, for the best development experience. In this case, I'm not going to just merge this "zero impact" change just to fix platforms we don't target.

I was not sure this was a 21.10 issue - this is running scons -u in the cereal folder, not running the full OP scons.

I will spin up a 20.04 and see if the issue reproduces. If so, how would you prefer to resolve? It sounds like you don't like attributes, so should I update scons to ignore the warning (which is treated as an error)?

Thanks!

JasonJShuler:
I have confirmed 20.04 doesn't display the gcc error. Closing PR.

PR #263, opened 2022-03-21
Add hondaRadarless SafetyModel

royjr:
Add hondaRadarless SafetyModel for radarless Honda cars like the 2022 Civic. Duplicate of #234 to retest failed check.

### DIFF
--- a/car.capnp
+++ b/car.capnp
@@ -546,6 +546,7 @@ struct CarParams {
     hyundaiCommunity @24;
     stellantis @25;
     faw @26;
+    hondaRadarless @27;
   }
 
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
No need to open a PR to run CI. If you enable GitHub Actions on your fork, the tests should just work.

royjr:
Thanks, it passed, Shall I resubmit PR?

adeebshihadeh:
Nope, there's already one open. I'd like to wait for the panda PR before deciding whether we actually need a new safety mode.

royjr:
Makes sense.

PR #269, opened 2022-03-29
car events: event for pre-enabling on brake

sshane:
`gasPressed` is renamed to `pedalPressedPreEnable`


sshane:
https://github.com/commaai/cereal/pull/268/files pretty much does this now

PR #281, opened 2022-04-14
set testJoystick frequency

adeebshihadeh:
Should be a good value for minimum frequency. needs joystick updates too


haraschax:
Doesn't seem relevant or urgent anymore.

PR #284, opened 2022-04-24
fixed set but not used build error. Removed apt '/' continuation.

MoreTore:
- Created new function to fix warning.
- removed apt continuation since this will be an error soon


Dockerfile
```diff
@@ -29,7 +29,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     tk-dev \
     wget \
     xz-utils \
-    zlib1g-dev \
+    zlib1g-dev
```
incognitojam, line 32: this was correct as it was

PR #290, opened 2022-04-30
body: add batteryPercent to car

kfatehi:
Adding batteryPercent to car in order to access its value from Qt UI

Related to https://github.com/commaai/openpilot/pull/24383

### DIFF
--- a/car.capnp
+++ b/car.capnp
@@ -205,6 +205,9 @@ struct CarState {
   leftBlindspot @33 :Bool; # Is there something blocking the left lane change
   rightBlindspot @34 :Bool; # Is there something blocking the right lane change
 
+  # body
+  batteryPercent @41 :UInt8;
+
   struct WheelSpeeds {
     # optional wheel speeds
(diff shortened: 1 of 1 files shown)


geohot:
A few things:
* why is it a uint8? use Float32.
* this should apply to cars as well, in gas car it's fuel in tank and electric it's battery

kfatehi:
> A few things:
> 
> * why is it a uint8? use Float32 (ranging from 0.0-1.0?)

No particular reason

> * this should apply to cars as well, in gas car it's fuel in tank and electric it's battery

Yeah I was thinking about this and considered calling it something like fuelLevel for both gas and electric. This probably makes sense considering we call these LiPo battery monitors "fuel gauges" for example the [MAX17043](https://www.maximintegrated.com/en/products/power/battery-management/MAX17043.html). I was surprised not to see an appropriate existing attribute to use and learned that it appears we don't track fuel or battery of the cars yet until the body appeared.

PR #291, opened 2022-04-30
add fuelGaugePercentage to CarState

kfatehi:
Replaces https://github.com/commaai/cereal/pull/290

A Float32 "fuelGaugePercentage" should be easy to understand as the
attribute to use to represent the fuel gauge value as a percentage
from 0.0 to 1.0. It should be generic enough for whatever power
source we call fuel be it gas, electric or something else.


geohot:
is it a Percentage now?

kfatehi:
sounds good https://github.com/commaai/cereal/pull/292

PR #299, opened 2022-05-10
messaging: support OPENPILOT_PREFIX

lukaspetersson:
Add a prefixing ID in the file structure path to messages.

Tested by:
* running test_processes
* running on device
* running latency logger on produced logs


messaging/messaging.cc
```diff
@@ -15,6 +17,7 @@ bool messaging_use_zmq(){
 Context * Context::create(){
   Context * c;
   if (messaging_use_zmq()){
+    assert(!std::getenv("OPENPILOT_PREFIX"));
```
adeebshihadeh, line 20: let's put this in `messaging_use_zmq` and print an error

messaging/msgq.cc
```diff
@@ -82,22 +83,26 @@ void msgq_wait_for_subscriber(msgq_queue_t *q){
 }
 
 
+std::string env_or_default(const std::string& variable_name) {
+  const char* value = std::getenv(variable_name.c_str());
+  return value ? value : "DEFAULT";
+}
```
adeebshihadeh, line 89: don't think a helper is necessary here

messaging/msgq.cc
```diff
@@ -82,22 +83,26 @@ void msgq_wait_for_subscriber(msgq_queue_t *q){
...
 
-  const char * prefix = "/dev/shm/";
-  char * full_path = new char[strlen(path) + strlen(prefix) + 1];
-  strcpy(full_path, prefix);
-  strcat(full_path, path);
+  std::string full_path = "/dev/shm/";
+  mkdir(full_path.c_str(), 0775);
+  full_path += env_or_default("OPENPILOT_PREFIX") + "/";
+  mkdir(full_path.c_str(), 0775);
+  full_path += path;
```
adeebshihadeh, line 99: let's make the directory in the helper that setups the prefixed environment since that's what openpilot does for `/dev/shm`
```suggestion
  const char* prefix = std::getenv("OPENPILOT_PREFIX");
  if (prefix) {
    full_path += std::string(prefix);
  }
```

messaging/messaging.cc
```diff
@@ -9,7 +12,13 @@ const bool MUST_USE_ZMQ = false;
 #endif
 
 bool messaging_use_zmq(){
-  return std::getenv("ZMQ") || MUST_USE_ZMQ;
+  if (std::getenv("ZMQ") || MUST_USE_ZMQ) {
+    if (std::getenv("OPENPILOT_PREFIX")) {
+      throw std::invalid_argument("OPENPILOT_PREFIX not supported with ZMQ backend");
```
adeebshihadeh, line 17:
```suggestion
      std::cerr << "OPENPILOT_PREFIX not supported with ZMQ backend\n";
      assert(false);
```

PR #301, opened 2022-05-11
Add desired curvatures to controlsState

JasonJShuler:
Adds the actual curvature and curvature rate that comes out of "get_lag_adjusted_curvature()" and is fed into and used by the lateral controllers. The fields are referred to in controlsd (and the parameter name in update() in the lateral controllers) as desired_curvature and desired_curvature_rate.

Added the fields to ControlsState.

This PR goes together with OP PR: https://github.com/commaai/openpilot/pull/24505

**Background**

While attempting to diagnose unusual tuning behavior in GM Trucks, I had a need to analyze the curvature, and discovered that the curvature actually used by the lateral controllers was not logged anywhere. I added it temporarily, and was surprised to discover that the output of get_lag_adjusted_curvature was in part responsible for the delayed and jerky steering.

### DIFF
--- a/log.capnp
+++ b/log.capnp
@@ -583,6 +583,10 @@ struct ControlsState @0x97ff69c53601abf1 {
   cumLagMs @15 :Float32;
   canErrorCounter @57 :UInt32;
 
+  # These are the adjusted curvatures that are actually passed to the lateral controllers
+  desiredCurvature @61 :Float32; # Lag adjusted curvature used by lateral controllers
+  desiredCurvatureRate @62 :Float32; # Lag adjusted curvature rate used by lateral controllers
+
   lateralControlState :union {
     indiState @52 :LateralINDIState;
     pidState @53 :LateralPIDState;
--- a/log.capnp
+++ b/log.capnp
@@ -568,6 +568,9 @@ struct ControlsState @0x97ff69c53601abf1 {
   ufAccelCmd @33 :Float32;
   aTarget @35 :Float32;
   curvature @37 :Float32;  # path curvature from vehicle model
+  # These are the adjusted curvatures that are actually passed to the lateral controllers
+  desiredCurvature @61 :Float32; # Lag adjusted curvature used by lateral controllers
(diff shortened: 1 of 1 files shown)


pd0wm:
I prefer putting them in controlsState like in this PR: https://github.com/commaai/openpilot/pull/24411

JasonJShuler:
> I prefer putting them in controlsState like in this PR: [commaai/openpilot#24411](https://github.com/commaai/openpilot/pull/24411)

Updated.

log.capnp
```diff
@@ -583,6 +583,10 @@ struct ControlsState @0x97ff69c53601abf1 {
   cumLagMs @15 :Float32;
   canErrorCounter @57 :UInt32;
 
+  # These are the adjusted curvatures that are actually passed to the lateral controllers
+  desiredCurvature @61 :Float32; # Lag adjusted curvature used by lateral controllers
+  desiredCurvatureRate @62 :Float32; # Lag adjusted curvature rate used by lateral controllers
```
sshane, line 588: put near curvature, above?

JasonJShuler (commented):
As requested, moved them to right below curvature

log.capnp
JasonJShuler: As requested, moved them to right below curvature

log.capnp
```diff
@@ -568,6 +568,9 @@ struct ControlsState @0x97ff69c53601abf1 {
   ufAccelCmd @33 :Float32;
   aTarget @35 :Float32;
   curvature @37 :Float32;  # path curvature from vehicle model
+  # These are the adjusted curvatures that are actually passed to the lateral controllers
+  desiredCurvature @61 :Float32; # Lag adjusted curvature used by lateral controllers
+  desiredCurvatureRate @62 :Float32; # Lag adjusted curvature rate used by lateral controllers
```
sshane, line 573:
```suggestion
  desiredCurvature @61 :Float32;  # lag adjusted curvatures used by lateral controllers
  desiredCurvatureRate @62 :Float32;
```

sshane:
Can you apply the suggestion? Don't have access to opgm

PR #307, opened 2022-05-20
SubMaster: add poll argument to C++ class

sshane:


messaging/socketmaster.cc
```diff
@@ -51,26 +51,29 @@ struct SubMaster::SubMessage {
...
+SubMaster::SubMaster(const std::vector<const char *> &service_list, const std::vector<const char *> &poll,
+                     const char *address, const std::vector<const char *> &ignore_alive) {
   poller_ = Poller::create();
   for (auto name : service_list) {
     const service *serv = get_service(name);
     assert(serv != nullptr);
     SubSocket *socket = SubSocket::create(message_context.context(), name, address ? address : "127.0.0.1", true);
     assert(socket != 0);
-    poller_->registerSocket(socket);
+    bool is_polled = inList(poll, name);
```
adeebshihadeh, line 68: default of empty vector should mean poll on all
```suggestion
    bool is_polled = poll.empty() || inList(poll, name);
```

messaging/socketmaster.cc
```diff
@@ -51,26 +51,29 @@ struct SubMaster::SubMessage {
...
+    bool is_polled = inList(poll, name) || poll.empty();
+    if (is_polled) poller_->registerSocket(socket);
     SubMessage *m = new SubMessage{
       .name = name,
       .socket = socket,
       .freq = serv->frequency,
       .ignore_alive = inList(ignore_alive, name),
-      .allocated_msg_reader = malloc(sizeof(capnp::FlatArrayMessageReader))};
+      .allocated_msg_reader = malloc(sizeof(capnp::FlatArrayMessageReader)),
+      .is_polled = inList(poll, name)};
```
adeebshihadeh, line 76: this isn't the same as is_polled declared above, why not just use it?

sshane, line 76:
```suggestion
      .is_polled = is_polled};
```

messaging/socketmaster.cc
```diff
@@ -81,6 +84,14 @@ void SubMaster::update(int timeout) {
   for (auto &kv : messages_) kv.second->updated = false;
 
   auto sockets = poller_->poll(timeout);
+
+  // non-blocking receive for non-polled sockets
```
sshane, line 88:
```suggestion
  // add non-polled sockets for non-blocking receive
```

PR #309, opened 2022-06-01
dcam is always 20Hz

pd0wm:


services.py
```diff
@@ -19,7 +17,8 @@ def __init__(self, port: int, should_log: bool, frequency: float, decimation: Op
     self.frequency = frequency
     self.decimation = decimation
 
-DCAM_FREQ = 10. if not TICI else 20.
+
+DCAM_FREQ = 20.
```
adeebshihadeh, line 21: remove this entirely? we don't do it for the other cams

PR #310, opened 2022-06-01
visionbuf use EGL texture

jwooning:
for https://github.com/commaai/openpilot/pull/24700

### DIFF
--- a/visionipc/visionbuf_ion.cc
+++ b/visionipc/visionbuf_ion.cc
@@ -12,11 +12,6 @@
 #include <linux/ion.h>
 #include <CL/cl_ext.h>
 
-#define EGL_EGLEXT_PROTOTYPES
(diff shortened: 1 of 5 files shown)


visionipc/visionbuf_ion.cc
```diff
@@ -117,6 +117,27 @@ void VisionBuf::init_cl(cl_device_id device_id, cl_context ctx) {
   assert(err == 0);
 }
 
+void VisionBuf::init_gl() {
```
adeebshihadeh, line 120: any reason to do this here instead of UI?

jwooning: the image/texture is linked to the shared buffers, so it requires a texture pointer for every visionbuf object, so I guess it makes more sense to put it there

PR #315, opened 2022-06-07
driverStateV2

adeebshihadeh:


log.capnp
```diff
@@ -1614,7 +1614,36 @@ struct Joystick {
...
+    facePositionStd @3 :List(Float32);
+    faceProb @4 :Float32;
+    leftEyeProb @5 :Float32;
+    rightEyeProb @6 :Float32;
+    leftBlinkProb @7 :Float32;
+    rightBlinkProb @8 :Float32;
+    sunglassesProb @9 :Float32;
+    occludedProb @10 :Float32;
+    readyProb @11 :List(Float32);
+    notReadyProb @12 :List(Float32);
```
adeebshihadeh, line 1642: notReadyProb isn't 1 - readyProb?

log.capnp
```diff
@@ -1614,7 +1614,36 @@ struct Joystick {
...
+  dspExecutionTime @2 :Float32;
+  rawPredictions @3 :Data;
+
+  poorVision @4 :Float32;
+  wheelOnRight @5 :Float32;
+
+  driverDataLH @6 :DriverData;
+  driverDataRH @7 :DriverData;
+
+  struct DriverData {
```
adeebshihadeh, line 1629: should we come up with a better name than driver?

log.capnp
```diff
@@ -1614,7 +1614,36 @@ struct Joystick {
...
 
-struct DriverState {
+struct DriverStateV2 {
+  frameId @0 :UInt32;
+  modelExecutionTime @1 :Float32;
+  dspExecutionTime @2 :Float32;
+  rawPredictions @3 :Data;
+
+  poorVision @4 :Float32;
+  wheelOnRight @5 :Float32;
```
adeebshihadeh, line 1624: I assume this is a probability? add the Prob suffix for consistency?

log.capnp
ZwX1616: no they don't mean the same thing

log.capnp
ZwX1616: can use "occupant" instead? if that doesn't sound weird

log.capnp
ZwX1616: yeah

PR #322, opened 2022-06-14
Allow cereal to build independently

jakepoz:
A few warnings in msgq.cc prevented cereal for building independently from openpilot.

- I updated the warning flags in SConstruct to match openpilot
- I applied an unused attribute to the unused variables to suppress the warning regardless


jakepoz:
Just noticed, https://github.com/commaai/cereal/pull/284 is probably a cleaner fix for this.

PR #339, opened 2022-07-29
tests action: checkout@v3 & made easier for forks

jeroenlammersma:
* Added DOCKER_REGISTRY to env variables.

* Updated checkout to v3.

* Use repository_owner variable for github repository


jeroenlammersma:
It fails, because I am not able to push to dockerhub.

PR #342, opened 2022-08-03
Additional longitudinal override event

ntegan1:
Adds a `brakePressedOverride` event so that forks can begin using common longitudinal override code introduced in #268.


adeebshihadeh:
We don't tend to merge additions only used in forks, but capnproto makes it pretty easy to extend the schema in your fork.

PR #349, opened 2022-08-30
CarParams: add field for auto resume capability

sshane:


car.capnp
```diff
@@ -424,6 +424,7 @@ struct CarParams {
   safetyConfigs @62 :List(SafetyConfig);
   alternativeExperience @65 :Int16;      # panda flag for features like no disengage on gas
   maxLateralAccel @68 :Float32;
+  autoResumeSng @69 :Bool;
```
adeebshihadeh, line 427: if this is just used for docs, can you add a comment? we should be a bit better about documenting our structs

sshane, line 427:
```suggestion
  autoResumeSng @69 :Bool;  # describes whether car can resume from a stop automatically
```

adeebshihadeh: also add something saying it's only used for docs

PR #357, opened 2022-09-09
Sensor events splitup

gast04:
split up sensorEvents and give every sensor its own service

### DIFF
--- a/log.capnp
+++ b/log.capnp
@@ -1948,7 +1948,12 @@ struct Event {
     gpsNMEA @3 :GPSNMEAData;
     can @5 :List(CanData);
     controlsState @7 :ControlsState;
-    sensorEvents @11 :List(SensorEventData);
+    sensorEvent @99 :SensorEventData;
+    gyroscope @94 :SensorEventData;
+    accelerometer @95 :SensorEventData;
+    magnetometer @96 :SensorEventData;
+    lightSensor @97 :SensorEventData;
+    temperatureSensor @98 :SensorEventData;
     pandaStates @81 :List(PandaState);
     peripheralState @80 :PeripheralState;
     radarState @13 :RadarState;
--- a/log.capnp
+++ b/log.capnp
@@ -1948,7 +1948,7 @@ struct Event {
     gpsNMEA @3 :GPSNMEAData;
     can @5 :List(CanData);
     controlsState @7 :ControlsState;
-    sensorEvents @11 :List(SensorEventData);
+    sensorEventsDEPRECATED @11 :List(SensorEventData);
     gyroscope @99 :SensorEventData;
     gyroscope2 @100 :SensorEventData;
     accelerometer @98 :SensorEventData;
--- a/log.capnp
+++ b/log.capnp
@@ -1949,7 +1949,7 @@ struct Event {
     can @5 :List(CanData);
     controlsState @7 :ControlsState;
     gyroscope @11 :SensorEventData;
-    accelerometer @94 :SensorEventData;
+    accelerometer @98 :SensorEventData;
     magnetometer @95 :SensorEventData;
     lightSensor @96 :SensorEventData;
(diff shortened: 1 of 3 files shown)


log.capnp
```diff
@@ -1910,7 +1910,11 @@ struct Event {
     gpsNMEA @3 :GPSNMEAData;
     can @5 :List(CanData);
     controlsState @7 :ControlsState;
-    sensorEvents @11 :List(SensorEventData);
```
adeebshihadeh, line 1913: in order to keep backwards compatibility with old logs, we don't remove fields, but add `DEPRECATED` at the end of the name and move it to the bottom of the struct with the other deprecated fields

services.py
```diff
@@ -21,7 +21,11 @@ def __init__(self, port: int, should_log: bool, frequency: float, decimation: Op
 services = {
   # service: (should_log, frequency, qlog decimation (optional))
   # note: the "EncodeIdx" packets will still be in the log
-  "sensorEvents": (True, 100., 100),
+  "gyroscope": (True, 100., 100),
+  "accelerometer": (True, 100., 100),
```
adeebshihadeh, line 25: aren't these 104Hz?

log.capnp
```diff
@@ -1949,6 +1949,13 @@ struct Event {
     can @5 :List(CanData);
     controlsState @7 :ControlsState;
     sensorEvents @11 :List(SensorEventData);
```
adeebshihadeh, line 1951: add the DEPRECATED suffix and move it down with the other deprecated events
```suggestion
    sensorEventsDEPRECATED @11 :List(SensorEventData);
```

services.py
```diff
@@ -22,6 +22,13 @@ def __init__(self, port: int, should_log: bool, frequency: float, decimation: Op
   # service: (should_log, frequency, qlog decimation (optional))
   # note: the "EncodeIdx" packets will still be in the log
   "sensorEvents": (True, 100., 100),
+  "gyroscope": (True, 100., 100),
```
adeebshihadeh, line 25: aren’t some of these 104Hz?

gast04: yes, gyro and accel are 104Hz, just thought as we in general aim for 100Hz, changed

log.capnp
gast04: done

services.py
```diff
@@ -21,7 +21,14 @@ def __init__(self, port: int, should_log: bool, frequency: float, decimation: Op
 services = {
   # service: (should_log, frequency, qlog decimation (optional))
   # note: the "EncodeIdx" packets will still be in the log
-  "sensorEvents": (True, 100., 100),
+  "sensorEventsDEPRECATED": (True, 100., 100),
```
adeebshihadeh, line 24: deprecated services don't need to be here, just need to be kept in the cereal structs

services.py
```diff
@@ -21,7 +21,14 @@ def __init__(self, port: int, should_log: bool, frequency: float, decimation: Op
...
+  "gyroscope": (True, 104., 100),
+  "gyroscope2": (True, 100., 100),
+  "accelerometer": (True, 104., 100),
```
adeebshihadeh, line 27: update the qlog decimation to be 104 as well

services.py
gast04: updated

services.py
gast04: updated

adeebshihadeh (approved):

services.py
```diff
@@ -21,7 +21,13 @@ def __init__(self, port: int, should_log: bool, frequency: float, decimation: Op
 services = {
   # service: (should_log, frequency, qlog decimation (optional))
   # note: the "EncodeIdx" packets will still be in the log
-  "sensorEvents": (True, 100., 100),
```
adeebshihadeh, line 24: also need to add this back

PR #358, opened 2022-09-09
add RP with chiplet

briskspirit:


log.capnp
```diff
@@ -455,6 +455,7 @@ struct PandaState @0xa7649e2575e4591e {
     uno @5;
     dos @6;
     redPanda @7;
+    red2Panda @8;
```
adeebshihadeh, line 458: think this is a little more visually distinct
```suggestion
    redPandaV2 @8;
```

PR #361, opened 2022-09-15
Add panda CAN health to logs

briskspirit:


log.capnp
```diff
@@ -474,6 +474,36 @@ struct PandaState @0xa7649e2575e4591e {
   safetyParam2DEPRECATED @26 :UInt32;
 }
 
+struct PandaCanState {
```
adeebshihadeh, line 477: would probably put this inside the PandaState struct

PR #363, opened 2022-09-23
bridge: only republish services that have subscribers

deanlee:
resolve https://github.com/commaai/cereal/issues/282
the bridge will start pushing messages if 'managerState' has been subscribed.


deanlee:
@adeebshihadeh do we need to exit/restart bridge if client disconnected?

PR #366, opened 2022-09-24
msgq_poll:  wait on signal instead of sleep for a fixed time,return immediately on new msg

deanlee:
msgq_poll can work as expected now. it return immediately on new msg or timeout.


deanlee:
The python version of the submaser should be modified too, we can safely register all sockets for msgq_poll

deanlee:
moved to https://github.com/commaai/cereal/pull/458

PR #374, opened 2022-10-11
CarParams cleanup

adeebshihadeh:


car.capnp
```diff
@@ -406,23 +406,27 @@ struct CarControl {
 # ****** car param ******
 
 struct CarParams {
+  # fingerprinting and identification
   carName @0 :Text;
```
incognitojam, line 410: `brandName`?

incognitojam, line 410: `brandName`?

PR #375, opened 2022-10-15
Add temporarily ACC fault event

sshane:


car.capnp
```diff
@@ -66,6 +66,7 @@ struct CarEvent @0x9b1657f34caf3ad3 {
     lowBattery @48;
     vehicleModelInvalid @50;
     accFaulted @51;
+    gmAccFaultedTemp @115;
```
sshane, line 69: We can rename to accFaultedTemp if we end up using this for other makes

jyoung8607: This would be useful for VW, we explicitly have both temp and perm faults.

sshane: Cool, open a PR to rename it when you want to use it. In GM's case it's to prevent an ACC fault from happening in certain situations

PR #381, opened 2022-11-02
add curvature steer control type

incognitojam:
- https://github.com/commaai/openpilot/pull/26236

### DIFF
--- a/car.capnp
+++ b/car.capnp
@@ -335,6 +335,7 @@ struct CarControl {
     # range from -1.0 - 1.0
     steer @2: Float32;
     steeringAngleDeg @3: Float32;
+    curvature @7: Float32;
 
     speed @6: Float32; # m/s
     accel @4: Float32; # m/s^2
@@ -346,7 +347,6 @@ struct CarControl {
       stopping @2;
       starting @3;
     }
-
   }
 
   struct CruiseControl {
@@ -586,6 +586,7 @@ struct CarParams {
   enum SteerControlType {
     torque @0;
     angle @1;
+    curvature @2;
   }
 
   enum TransmissionType {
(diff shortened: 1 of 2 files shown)


log.capnp
```diff
@@ -742,6 +743,19 @@ struct ControlsState @0x97ff69c53601abf1 {
...
+    active @0 :Bool;
+    actualCurvature @1 :Float32;
+    desiredCurvature @2 :Float32;
+    error @3 :Float32;
+    p @4 :Float32;
+    i @5 :Float32;
+    d @6 :Float32;
+    f @7 :Float32;
+    output @8 :Float32;
+    saturated @9 :Bool;
```
incognitojam, line 756: maybe steeringAngleDeg and steeringAngleDesiredDeg would be good here too?

car.capnp
```diff
@@ -335,6 +335,7 @@ struct CarControl {
     # range from -1.0 - 1.0
     steer @2: Float32;
     steeringAngleDeg @3: Float32;
+    curvature @7: Float32;
```
adeebshihadeh, line 338: add a new line so it's not grouped with steer angle for the comment above
```suggestion

    curvature @7: Float32;
```

log.capnp
```diff
@@ -742,6 +743,19 @@ struct ControlsState @0x97ff69c53601abf1 {
...
   }
 
+  struct LateralCurvatureState {
+    active @0 :Bool;
+    actualCurvature @1 :Float32;
+    desiredCurvature @2 :Float32;
+    error @3 :Float32;
+    p @4 :Float32;
+    i @5 :Float32;
+    d @6 :Float32;
```
adeebshihadeh, line 753: are you using any derivative?

adeebshihadeh (approved):

log.capnp
incognitojam: Not right now, and curvature controller as is doesn't calculate the rate so doesn't support it

PR #382, opened 2022-11-06
add microphone

incognitojam:


log.capnp
```diff
@@ -1961,6 +1961,11 @@ struct EncodeData {
 struct UserFlag {
 }
 
+struct Microphone {
+  noiseLevel @0 :Float32;
+  noiseLevelRaw @1 :Float32;
```
adeebshihadeh, line 1966: would add "ambient" prefix and "db" suffix

adeebshihadeh: also add a comment saying the other one is filtered

adeebshihadeh (approved):
Feel free to merge after addressing comments.

log.capnp
incognitojam: I'm not sure it actually is dB

PR #400, opened 2022-12-30
vipc: add function to get available streams

deanlee:


visionipc/visionipc_tests.cc
```diff
@@ -22,6 +22,17 @@ TEST_CASE("Connecting"){
...
 
+TEST_CASE("getAvailableStreams"){
+  VisionIpcServer server("camerad");
+  server.create_buffers(VISION_STREAM_ROAD, 1, false, 100, 100);
+  server.create_buffers(VISION_STREAM_WIDE_ROAD, 1, false, 100, 100);
+  server.start_listener();
+  auto available_streams = VisionIpcClient::getAvailableStreams("camerad");
+  REQUIRE(available_streams.size() == 2);
+  REQUIRE(available_streams[0] == VISION_STREAM_ROAD);
+  REQUIRE(available_streams[1] == VISION_STREAM_WIDE_ROAD);
```
adeebshihadeh, line 33: does it make sense to return an std::set instead to better write things like this?

visionipc/visionipc_client.cc
```diff
@@ -29,23 +40,10 @@ bool VisionIpcClient::connect(bool blocking){
...
-    if (socket_fd < 0) {
-      if (blocking){
-        std::cout << "VisionIpcClient connecting" << std::endl;
-        std::this_thread::sleep_for(std::chrono::milliseconds(100));
-      } else {
-        return false;
-      }
-    }
+  int socket_fd = connect_to_vipc_server(name, blocking);
+  if (socket_fd == -1) {
```
adeebshihadeh, line 44:
```suggestion
  if (socket_fd < 0) {
```

PR #410, opened 2023-02-04
added smartdsu as a CarParams

eFiniLan:
As per PR in openpilot: https://github.com/commaai/openpilot/pull/27212
Passing smartDsu value over to carstate so we can stop "ACC_CONTROL" being read.

### DIFF
--- a/car.capnp
+++ b/car.capnp
@@ -424,6 +424,7 @@ struct CarParams {
   pcmCruise @3 :Bool;        # is openpilot's state tied to the PCM's cruise state?
   enableDsu @5 :Bool;        # driving support unit
   enableBsm @56 :Bool;       # blind spot monitoring
+  enableSmartDsu @72 :Bool;  # smart dsu
   flags @64 :UInt32;         # flags for car specific quirks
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
We're moving away from adding vehicle-specific flags to carState. Instead, carState has a [flags field](https://github.com/commaai/cereal/blob/bdbac40160a550c88e481a3bf700a1016901bd9b/car.capnp#L427) that you can use [like this](https://github.com/commaai/openpilot/blob/f5f14068173fbf8664e682e5b02270ecbc79abdb/selfdrive/car/hyundai/values.py#L54).

eFiniLan:
> We're moving away from adding vehicle-specific flags to carState. Instead, carState has a [flags field](https://github.com/commaai/cereal/blob/bdbac40160a550c88e481a3bf700a1016901bd9b/car.capnp#L427) that you can use [like this](https://github.com/commaai/openpilot/blob/f5f14068173fbf8664e682e5b02270ecbc79abdb/selfdrive/car/hyundai/values.py#L54).

Cool, thanks!

PR #414, opened 2023-03-02
add vAbs in RadarPoint

Smartype:
Toyota CAMRY/CHR radar report absolute velocities of targets. Add this field so that we do not have to touch radar_interface of other cars


adeebshihadeh:
Can you just subtract vEgo instead?

PR #422, opened 2023-03-19
VW: Add CarEvent for EPS timer soft disable

jyoung8607:


car.capnp
```diff
@@ -114,6 +114,7 @@ struct CarEvent @0x9b1657f34caf3ad3 {
     canBusMissing @111;
     controlsdLagging @112;
     resumeBlocked @113;
+    steerTimeLimit @116;
```
sshane, line 117: Maybe you an re-use accFaultedTemp since we never really used it?

jyoung8607: Done.

PR #426, opened 2023-03-24
Ecu: add parking

sshane:


car.capnp
```diff
@@ -643,6 +643,9 @@ struct CarParams {
     # Chrysler only
     hcp @18;  # Hybrid Control Processor
 
+    # Hyundai only
+    parking @22;  # ADAS parking ECU
```
adeebshihadeh, line 647: this isn't really hyundai specific. it's the equivalent of toyota's `apgs`

sshane: i'll rename to parking then and use @22 for another ECU later. there can be a few different ecu abreviations; ipas, apgs (which is lexus only), rspa (hyundai).

PR #430, opened 2023-03-30
deprecate liveParameters.yawRate

adeebshihadeh:


nuwandavek:
Can we wait on this? The RAM steering angle sensor is quite bad (laggy), so we might have to use the localizer yaw rate to close the loop. And the `liveparameters` yaw rate is a smoothed version of that, which might be better to use.

adeebshihadeh:
We can revert once we're ready to use it.

PR #439, opened 2023-05-02
FakeSubSocket and FakePubSocket for IPC synchronization

fredyshox:
General usage flow showcased in fake_demo.cc

Fake sockets (which are wrappers MSGQ/ZMQ sockets) can be enabled with `CEREAL_FAKE` env var. File descriptors for events are currently passed via ugly env vars, `CEREAL_FAKE_RECV_CALLED_FD_[endpoint]` and `CEREAL_FAKE_RECV_READY_FD_[endpoint]`. If those env vars are not set, the sockets work the same as their underlying implementations. 

This is completely invisible from the daemon perspective - no changes needed. The testing procedure may use `FakeEvent` API as a way of controlling synchronization.


fredyshox:
process_replay seems to be working for all the processes, so we can work towards merging this first, then the openpilot changes

messaging/event.cc
```diff
@@ -0,0 +1,181 @@
...
+int Event::fd() {
+  return event_fd;
+}
+
+Event * Event::create(std::string endpoint, EventPurpose purpose) {
+  int fd;
+
+  if (!event_fd_from_environ(endpoint, purpose, &fd)) {
+    return nullptr;
+  }
```
adeebshihadeh, line 120: these can all be asserts

adeebshihadeh:
We'll also need to adjust this to work with OP in the pipeline, since this relies on the PubSocket being in the replay process.

fredyshox:
I've changed the API a little bit, right now the user is not creating events directly. Instead they use `SocketEventHandle` which wraps recv_called and recv_ready events, but also can enable/disable them all will. Works in coordination with child process FakeSubSockets via shared memory.

PR #447, opened 2023-05-17
tests: fix recv_one_or_none test

fredyshox:


messaging/tests/test_messaging.py
```diff
@@ -191,17 +191,17 @@ def test_recv_one(self):
   def test_recv_one_or_none(self):
```
adeebshihadeh, line 191: let's add an assert raises to handle the ZMQ case so we can get this merged

fredyshox: hmm, im not sure if i get that, what do you mean? the easiest workaround to make it pass on ZMQ, that i can think of is to add `zmq_sleep` after send

adeebshihadeh: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

fredyshox: ======================================================================
FAIL: test_recv_one_or_none (messaging.tests.test_messaging.TestMessaging)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/project/cereal/messaging/tests/test_messaging.py", line 205, in test_recv_one_or_none
    self.assertIsInstance(recvd, capnp._DynamicStructReader)
AssertionError: None is not an instance of <class 'capnp.lib.capnp._DynamicStructReader'>
```

Comment (adeebshihadeh) [reply messaging/tests/test_messaging.py:191]: I just mean it should mark the ZMQ case as expected to fail, see https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures

File: messaging/tests/test_messaging.py [hunk 2]
```diff
@@ -188,20 +188,21 @@ def test_recv_one(self):
     self.assertIsInstance(recvd, capnp._DynamicStructReader)
     assert_carstate(msg.carState, recvd.carState)
 
+  unittest.skipIf("ZMQ" in os.environ, "send() is not blocking on ZMQ, causing test to fail")
```
adeebshihadeh, line 191: the test should still run, but expect failure if running with ZMQ. that way we know if the behavior changes

fredyshox: Since built-in `expectedFailure` decorator does not support conditions, I've added custom `zmq_expected_failure` which applies it only in ZMQ mode

adeebshihadeh (approved):

PR #451, opened 2023-05-24
Disable cereal fake APIs on macOS

fredyshox:
macOS lacks `sys/eventfd.h` and some other linux specific which are powering fake sockets. This is using compile-time definitions to remove event.cc and impl_fake.cc from being compiled, and hides python APIs related to that feature.

### DIFF
--- a/messaging/__init__.py
+++ b/messaging/__init__.py
@@ -1,8 +1,12 @@
 # must be build with scons
-from .messaging_pyx import Context, Poller, SubSocket, PubSocket, SocketEventHandle, toggle_fake_events, set_fake_prefix, get_fake_prefix, delete_fake_prefix, wait_for_one_event  # pylint: disable=no-name-in-module, import-error
-from .messaging_pyx import MultiplePublishersError, MessagingError  # pylint: disable=no-name-in-module, import-error
 import os
 import capnp
+import platform
+
+from .messaging_pyx import Context, Poller, SubSocket, PubSocket # pylint: disable=no-name-in-module, import-error
+from .messaging_pyx import MultiplePublishersError, MessagingError  # pylint: disable=no-name-in-module, import-error
+if platform.system() != "Darwin":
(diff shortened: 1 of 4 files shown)


messaging/messaging.cc
```diff
@@ -94,15 +104,19 @@ PubSocket * PubSocket::create(Context * context, std::string endpoint, bool chec
...
   } else {
+#endif
     if (messaging_use_zmq()){
       p = new ZMQPoller();
     } else {
       p = new MSGQPoller();
     }
+#ifdef CEREAL_FAKE
   }
+#endif
```
adeebshihadeh, line 119: this is not readable

fredyshox: Should be better now - made the conditional blocks more self-contained.

messaging/__init__.py
```diff
@@ -33,7 +38,8 @@
 context = Context()
 
 
-def fake_event_handle(endpoint: str, identifier: Optional[str] = None, override: bool = True, enable: bool = False) -> SocketEventHandle:
+def fake_event_handle(endpoint: str, identifier: Optional[str] = None, override: bool = True, enable: bool = False):
```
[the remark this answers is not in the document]
fredyshox: since this function will fail with an assert on macos anyway, should i just hide it, like the imports above?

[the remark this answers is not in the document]
fredyshox: I've changed it to export empty `SocketEventHandle` class on macOS just for typing purposes

messaging/__init__.py
```diff
@@ -12,11 +16,12 @@
 
 assert MultiplePublishersError
 assert MessagingError
-assert toggle_fake_events
-assert set_fake_prefix
-assert get_fake_prefix
-assert delete_fake_prefix
-assert wait_for_one_event
+if platform.system() != "Darwin":
```
adeebshihadeh, line 19: I think a stub implementation would be nicer here. This PR is +30 lines, which is fine, but it would be nicer if those 30 lines were in a single file with single `if mac`.  currently, there's a little bit of added code all of the place with a bunch of `if mac`

adeebshihadeh (changes requested):

fredyshox:
#476

PR #454, opened 2023-05-30
carState: non-critical car faulted field

sshane:
true if some ECU is faulted, but car remains controllable (for ex. camera faulted, but we are controlling both lat and long)

also for ex. camera faulted and we are not controlling long, that would be CRITICAL (accFaulted)

### DIFF
--- a/car.capnp
+++ b/car.capnp
@@ -116,7 +116,6 @@ struct CarEvent @0x9b1657f34caf3ad3 {
     resumeBlocked @113;
     steerTimeLimit @115;
     vehicleSensorsInvalid @116;
-    carFaultedNonCritical @118;
 
     radarCanErrorDEPRECATED @15;
     communityFeatureDisallowedDEPRECATED @62;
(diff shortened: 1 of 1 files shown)


car.capnp
```diff
@@ -116,6 +116,7 @@ struct CarEvent @0x9b1657f34caf3ad3 {
     resumeBlocked @113;
     steerTimeLimit @115;
     vehicleSensorsInvalid @116;
+    adasFaulted @118;
```
adeebshihadeh, line 119: it's not just adas faulted, it's anything faulted that doesn't go in accFaulted or steer fault

sshane: how about `carFaultedNonCritical`? I want it to be clear without a comment

car.capnp
```diff
@@ -166,7 +167,7 @@ struct CarState {
   # gas pedal, 0.0-1.0
   gas @3 :Float32;        # this is user pedal only
   gasPressed @4 :Bool;    # this is user pedal only
-  
+
```
sshane, line 170:
```suggestion
  
```

car.capnp
```diff
@@ -166,7 +166,7 @@ struct CarState {
...
```
sshane, line 169:
```suggestion
  
```

car.capnp
sshane: since we are choosing to not make it a noEntry, I'll leave it with this prefix. if there is another type of fault that is not acc or lkas in the future, but *does* disallow disengagement, we can just add `carFaulted`

PR #455, opened 2023-05-30
Deprecate brakeUnavailable event

sshane:
We can, in the future, revive this to distinguish ABS module/brake controller related faults and PCM/accelerator control faults in the qlogs, but for now, Honda was the only make using this (for both brake and normal PCM ACC faults), so remove.


car.capnp
```diff
@@ -166,7 +166,7 @@ struct CarState {
   # gas pedal, 0.0-1.0
   gas @3 :Float32;        # this is user pedal only
   gasPressed @4 :Bool;    # this is user pedal only
-  
+
```
sshane, line 169:
```suggestion
  
```

PR #456, opened 2023-06-01
connect: add error messages when Socket connect fails

fredyshox:
This aims to make socket create/connect failures more descriptive. Currently the c++ op services and Pub/SubMasters usually fail with assertion or segfault, which does not provide any information about the reason (which may be trivial to solve).

This happens very often with ZMQ, when some other service on your local machine is using some 80XX port.

This is also alignment with cython wrappers which do throw an error when connect fails.

### DIFF
--- a/messaging/messaging.cc
+++ b/messaging/messaging.cc
@@ -65,7 +65,9 @@ SubSocket * SubSocket::create(Context * context, std::string endpoint, std::stri
     return s;
   } else {
     delete s;
-    return NULL;
+    throw std::runtime_error(
+      "Failed to connect to " + endpoint + ": " + std::string(strerror(errno))
+    );
   }
 }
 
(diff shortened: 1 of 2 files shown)


messaging/messaging.cc
```diff
@@ -65,7 +65,9 @@ SubSocket * SubSocket::create(Context * context, std::string endpoint, std::stri
     return s;
   } else {
     delete s;
-    return NULL;
+    throw std::runtime_error(
+      "Failed to connect to " + endpoint + ": " + std::string(strerror(errno))
+    );
```
adeebshihadeh, line 70: we tend to avoid cpp exceptions. you can print to stderr instead

fredyshox: Ok, i guess that works too

PR #459, opened 2023-06-02
wake up readers in msgq_init_publisher

deanlee:
Wake up readers when initializing publisher, so that readers can update their read_uid in time.


adeebshihadeh:
Is this a bugfix? What does it fix?

deanlee:
this PR try to fixes an issue where subscribers may lose the first message in the new queue after the publisher is restarted for some reason.
It's not a big deal. and this pr is missing testcase, so I'll close it for now.

PR #463, opened 2023-06-07
visionipc: get_endpoint_name python binding

fredyshox:


adeebshihadeh:
What's the use case for this?

fredyshox:
support for modeld in process replay

PR #476, opened 2023-06-19
macOS: stub implementation of event API

fredyshox:


messaging/event.cc
```diff
@@ -6,17 +6,27 @@
...
 #include <sys/mman.h>
 #include <sys/stat.h>
 #include <unistd.h>
 #include <poll.h>
 #include <signal.h>
 #include <fcntl.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#ifdef __APPLE__
+#define ppoll(fds, nfds, timeout, sigmask) poll(fds, nfds, timeout != nullptr ? (timeout->tv_sec * 1000 + timeout->tv_nsec / 1000000) : -1)
```
adeebshihadeh, line 18: what's this for? wouldn't it hit an assert before this is even called?

messaging/event.cc
```diff
@@ -6,17 +6,27 @@
...
+#else
+#include <sys/eventfd.h>
+#endif 
 
 #include "cereal/messaging/event.h"
 
 void event_state_shm_mmap(std::string endpoint, std::string identifier, char **shm_mem, std::string *shm_path) {
+  #ifdef __APPLE__
+  std::cerr << "event_state_shm_mmap not supported on macOS" << std::endl;
+  assert(false);
```
adeebshihadeh, line 28: doesn't the single assert in the constructor cover it?

messaging/event.cc
```diff
@@ -6,17 +6,27 @@
...
-#include <sys/eventfd.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
 #include <unistd.h>
 #include <poll.h>
 #include <signal.h>
 #include <fcntl.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#ifdef __APPLE__
```
adeebshihadeh, line 17: add a new line here to break this up

messaging/event.cc
fredyshox: it would, but `wait`, `peek` and `wait_for_one` use that so i've added this stub for compilation to succeed

messaging/event.cc
fredyshox: it could, but thought about adding this check to prevent people from ever calling it, cause it uses the Linux-only shm fs

messaging/event.cc
fredyshox: otherwise would need couple of additional `#ifdef`s in the mentioned methods

messaging/event.cc
adeebshihadeh: let's remove this

messaging/event.cc
adeebshihadeh: can you try to write this more like this one? https://github.com/commaai/cereal/blob/master/logger/logger.h

fredyshox: Made it variadic macro to skip unused sigmask arg, not sure if that's what you meant

fredyshox: I think i get it know, take a look at `PPOLL` now

messaging/event.cc
```diff
@@ -206,3 +222,11 @@ int Event::wait_for_one(const std::vector<Event>& events, int timeout_sec) {
 
   throw std::runtime_error("Event poll failed, no events ready");
 }
+#else
+// Stub implementation for Darwin, which does not support eventfd
+void Event::set() const {}
```
adeebshihadeh, line 219: can you move this to the top?

PR #478, opened 2023-06-20
Added minimal set of tests for python VisionIpc

mitchellgoffpc:


visionipc/tests/test_visionipc.py
```diff
@@ -0,0 +1,71 @@
...
+  if "ZMQ" in os.environ:
+    time.sleep(t)
+
+
+class TestVisionIpc(unittest.TestCase):
+
+  def tearDown(self):
+    if hasattr(self, 'server'):
+      del self.server
+      gc.collect()
```
adeebshihadeh, line 19: why is this necessary?

mitchellgoffpc: I was seeing occasional `Error, failed to bind PubSocket to 9000: Address already in use` errors in CI, wanted to check if this fixed it

PR #485, opened 2023-06-22
CarParams.Ecu: rename hcp to hybrid

sshane:
Originally added for: https://github.com/commaai/openpilot/pull/25606. This makes it a generic Ecu we can use for different brands, specifically for Toyota here: https://github.com/commaai/openpilot/pull/28628.

Needs openpilot PR: https://github.com/commaai/openpilot/pull/28629

- Some info about the Chrysler HCP: https://www.allpar.com/threads/deep-inside-the-chrysler-pacifica-hybrid-system.230025/

  Summary: HCP has (is part of) an inverter and controls flow of energy and the motors.
- Some info about Toyota's hybrid system: https://en.wikipedia.org/wiki/Hybrid_Synergy_Drive

  Summary: Toyota also has a computer/ECU that manages the power flow between engine, MG2 (motor generator), and battery.
- Honda is similar with the [Integrated Motor Assist](https://en.wikipedia.org/wiki/Integrated_Motor_Assist).

### DIFF
--- a/car.capnp
+++ b/car.capnp
@@ -628,7 +628,7 @@ struct CarParams {
     engine @4;
     unknown @5;
     transmission @8; # Transmission Control Module
-    hybrid @18;  # hybrid control unit. e.g. Chrysler's Hybrid Control Processor, Toyota's hybrid control computer
+    hybrid @18; # hybrid control unit e.g. Chrysler's HCP, Toyota's hybrid control computer, Honda's IMA Control Unit
     srs @9; # airbag
     gateway @10; # can gateway
     hud @11; # heads up display
--- a/car.capnp
+++ b/car.capnp
@@ -628,7 +628,7 @@ struct CarParams {
     engine @4;
     unknown @5;
     transmission @8; # Transmission Control Module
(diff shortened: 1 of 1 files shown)


car.capnp
```diff
@@ -628,6 +628,7 @@ struct CarParams {
     engine @4;
     unknown @5;
     transmission @8; # Transmission Control Module
+    hybrid @18; # hybrid control unit e.g. Chrysler's HCP, Toyota's hybrid control computer, Honda's IMA Control Unit
```
sshane, line 631:
```suggestion
    hybrid @18; # hybrid control unit e.g. Chrysler's HCP, Honda's IMA Control Unit, Toyota's hybrid control computer
```

car.capnp
```diff
@@ -628,6 +628,7 @@ struct CarParams {
     engine @4;
     unknown @5;
     transmission @8; # Transmission Control Module
+    hybrid @18; # hybrid control unit e.g. Chrysler's HCP, Honda's IMA Control Unit, Toyota's hybrid control computer
```
sshane, line 631:
```suggestion
    hybrid @18; # hybrid control unit, e.g. Chrysler's HCP, Honda's IMA Control Unit, Toyota's hybrid control computer
```

PR #497, opened 2023-07-17
visionipc: support OPENPILOT_PREFIX

fredyshox:


visionipc/visionipc_client.cc
```diff
@@ -9,7 +9,12 @@
 #include "cereal/logger/logger.h"
 
 static int connect_to_vipc_server(const std::string &name, bool blocking) {
+  char* prefix = std::getenv("OPENPILOT_PREFIX");
   std::string path = "/tmp/visionipc_" + name;
+  if (prefix) {
+    path = path + "_" + prefix;
```
adeebshihadeh, line 15: should be prefix + path, with a default prefix of /tmp/

fredyshox: or maybe an intermediate directory of /tmp/ + prefix + '/visionipc_' + name

fredyshox:
Now it's using OPENPILOT_PREFIX as filename prefix. Intermediate directories would need to be created and cleaned afterwards.

PR #503, opened 2023-07-27
vpic_server: add support for creating buffers in multiple threads

deanlee:


adeebshihadeh:
Is this needed for anything?

deanlee:
I will close this PR for now. It is a prepreq for supports concurrent initialization of cameras to speed up camerad startup time. But it requires a lot of testing.

deanlee:
reopen. 

concurrent call to camera_init can reduce the startup time of camerad by 1-2 seconds.

PR #517, opened 2023-08-17
mypy: check untyped definitions

sshane:
plus some other warnings from openpilot


mypy.ini
```diff
@@ -0,0 +1,12 @@
+[mypy]
+; third-party packages
```
adeebshihadeh, line 2: we should put as much as possible in pyproject.toml, looks like mypy supports that

messaging/tests/test_messaging.py
```diff
@@ -175,7 +175,8 @@ def test_recv_sock(self):
     time.sleep(0.01)
     recvd = messaging.recv_sock(sub_sock)
     self.assertIsInstance(recvd, capnp._DynamicStructReader)
-    assert_carstate(msg.carState, recvd.carState)
+    # https://github.com/python/mypy/issues/13038
+    assert_carstate(msg.carState, recvd.carState)  # type: ignore[union-attr]
```
sshane, line 179: mypy doesn't look at unittest's asserts for type narrowing. we can do `if recvd is not None:`, but that is a similar solution to just ignoring it

messaging/tests/test_pub_sub_master.py
```diff
@@ -20,7 +21,7 @@ def test_init(self):
     sm = messaging.SubMaster(events)
     for p in [sm.updated, sm.rcv_time, sm.rcv_frame, sm.alive,
               sm.sock, sm.freq, sm.data, sm.logMonoTime, sm.valid]:
-      self.assertEqual(len(p), len(events))
+      self.assertEqual(len(cast(Sized, p)), len(events))
```
sshane, line 24: all these variables are sized, not sure what's going on here. if you type any one to `dict` in SubMaster it doesn't complain, but if you type them properly (`DIct[str, int]`) then it does. didn't want to go with the first as I'm not sure how it works

PR #526, opened 2023-08-30
Roundabout fix: Add maneuverDegrees to NavInstruction

jakethesnake420:
https://github.com/commaai/openpilot/pull/29708


sshane:
Let's fix in navd, a field only specific to one maneuver doesn't sound the cleanest

PR #537, opened 2023-09-25
Added reserved fields for body controls

mitchellgoffpc:


log.capnp
```diff
@@ -2265,6 +2265,10 @@ struct Event {
     livestreamWideRoadEncodeData @121 :EncodeData;
     livestreamDriverEncodeData @122 :EncodeData;
 
+    customReservedText0 @124 :Text;
```
adeebshihadeh, line 2268: would consider changing the text to data

PR #543, opened 2023-10-10
magnetometerCalibration: Structs for magd

nuwandavek:
openpilot PR: https://github.com/commaai/openpilot/pull/30216

### DIFF
--- a/log.capnp
+++ b/log.capnp
@@ -2012,15 +2012,24 @@ struct LiveTorqueParametersData {
 
 struct MagnetometerCalibrationData {
   calibrated @0 :Bool;
-  center @1 :XYData;
-  axisLengths @2 :XYData;
-  rotationAngle @3 :Float32;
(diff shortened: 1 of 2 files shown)


log.capnp
```diff
@@ -1996,6 +1996,16 @@ struct LiveTorqueParametersData {
   useParams @12 :Bool;
 }
 
+struct MagnetometerCalibrationData {
```
adeebshihadeh, line 1999: I think we want this in two parts: a `SensorEventData` for the calibrated measurements and perhaps another struct for the calibration data itself for debugging

nuwandavek: I think maybe not. The final "calibrated measurements" are not the only information needed to get bearing. There is an additional step to adjust for offsets in angle-domain. So the final post-all-calculations value would be the magnetometer bearing. But that's not a sensor data - so maybe it is better as is.

PR #546, opened 2023-10-17
generate python stubs for type completion

jnewb1:
example type completions:
image
image
image


adeebshihadeh:
There's a PR for this #493. This is way cleaner though; does it have the same coverage?

jnewb1:
> There's a PR for this #493. This is way cleaner though; does it have the same coverage?

coverage should be the same. requires changes to capnp-stub-generator to sepecify the output directory. enums are also broken because the current stubs use string literals, so something like

```car.CarParams.Ecu.fwdRadar```

doesn't work atm

PR #548, opened 2023-10-30
honda CAN FD

vanillagorillaa:


adeebshihadeh:
Is this for a whole new safety mode? If not, we’ll likely use a safety param instead.

PR #550, opened 2023-11-02
NLP driving - new model outputs

YassineYousfi:


adeebshihadeh:
How much does this increase qlog size?

YassineYousfi:
This adds 264 floats per message. 
It doesn't significantly increase the qlog file sizes.

PR #560, opened 2023-12-06
Chrysler: Add safety mode for Compact US Wide

jyoung8607:
Holding in draft pending decisions on how to factor this together with the existing Chrysler port.


jyoung8607:
Having reviewed both platforms side-by-side, CUSW is similar to Chrysler/Ram DT/Ram HD in a few ways, but not enough to sensibly merge CUSW support into the existing Chrysler safety. I'd like to proceed with a standalone CUSW safety ID. 

Ready for review.

PR #562, opened 2023-12-11
visionipc:  fix assert in getAvailableStreams

deanlee:
the vipc client still can connect to the ipc path even if the camerad is not running.   This could be due to the camerad not shutting down properly and the ipc path not being removed.

`ipc_sendrecv_with_fds` will returns -1 In this case. this causes the following assertion to fail.

> ui: cereal/visionipc/visionipc_client.cc:133: static std::set<VisionStreamType> VisionIpcClient::getAvailableStreams(const std::string &, bool): Assertion `(r >= 0) && (r % sizeof(VisionStreamType) == 0)' failed.

This bug can be reproduced by clicking “preview driver camera" button quickly and continuously in the UI interface.

### DIFF
--- a/visionipc/visionipc_client.cc
+++ b/visionipc/visionipc_client.cc
@@ -1,3 +1,4 @@
+#include <algorithm>
 #include <chrono>
 #include <cassert>
 #include <iostream>
--- a/visionipc/visionipc_client.cc
+++ b/visionipc/visionipc_client.cc
@@ -123,16 +123,22 @@ std::set<VisionStreamType> VisionIpcClient::getAvailableStreams(const std::strin
   if (socket_fd < 0) {
     return {};
   }
+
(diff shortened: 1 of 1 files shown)


deanlee:
VipcClient::recv and vipcClient::connect have similar problems, e.g. vipcClient:connect will returns true even  if an attempt is made to connect to a non-existent stream type.considering use another pr to fix these issues.

adeebshihadeh:
Doesn't the map still use it?

deanlee:
the map_render creates vipc buffer as yuv. Looks like a bug, though it doesn't really matter the result.
https://github.com/commaai/openpilot/blob/5f045176775c8c2cdd7f85e7229f78cf26f0c013/selfdrive/navd/map_renderer.cc#L92

deanlee:
closed , moved to https://github.com/commaai/msgq/pull/624

PR #572, opened 2024-01-18
fix logMonoTime, as c++ use boottime

Smartype:
This causes error in locationd on my Google Pixel 4. as calibrationd use mono time, which differs from others.

Observation timestamp is older than the max rewind threshold of the filter.


adeebshihadeh:
I assume this is because your Pixel will actually sleep? This will have to be fixed in quite a few more places.

Smartype:
> I assume this is because your Pixel will actually sleep? This will have to be fixed in quite a few more places.

Yes. As it is stock android 13.

Kernel rebuilt to enable USB host.

Rooted.

adeebshihadeh:
I'm supportive of fixing this, but I don't want to merge this change alone. It doesn't completely fix the problem and there is nothing preventing this from breaking again.

PR #573, opened 2024-01-21
Delete lateralPlan

haraschax:


log.capnp
```diff
@@ -1093,29 +1124,6 @@ struct LateralPlan @0xe1e9318e2ae8b51e {
     u @1 :List(Float32);
   }
 
-  enum Desire {
```
adeebshihadeh, line 1096: these can’t be deleted

PR #577, opened 2024-02-01
add secOCKey to carParams

pd0wm:
Probably useful to ensure routes can be used in process replay. Can also be fetched from initData if you don't want it in the carParams.


adeebshihadeh:
I don't think we want to log the keys at all. Any reason to besides offline tools/debugging?

pd0wm:
Without logging the key something like process replay wouldn’t work on SecOC routes. It will use the key to verify messages from the car, and throw an alert if you’re using the wrong key.

haraschax:
This is just a msgq library now

PR #589, opened 2024-02-15
Ecu: add vcu

sshane:
Some context about this ECU: https://github.com/commaai/openpilot/pull/25607#issuecomment-1239969101

On Hyundai it's 0x7e2 (not useful since it's not on OBD), and on the Nissan Leaf, it's 0x797 with an rx offset of 3, and has the VIN on bus 0.

Originally added here but replaced since we didn't use it: https://github.com/commaai/cereal/pull/348


sshane:
don't need FW version yet, can open if we need this from Nissan. generally isn't part of the API

PR #590, opened 2024-02-16
 paramsd: Add flag and alert for wheel misalignment

nuwandavek:
Ordinals necessary for https://github.com/commaai/openpilot/pull/31492


adeebshihadeh:
@haraschax this is pretty important to get back in. it pops up in support somewhat frequently

haraschax:
I think i just want to rewrite this entire thing. Looking into how to scope it out.

PR #610, opened 2024-05-09
card fields

sshane:


car.capnp
```diff
@@ -154,6 +154,7 @@ struct CarState {
   # CAN health
   canValid @26 :Bool;       # invalid counter/checksums
   canTimeout @40 :Bool;     # CAN bus dropped out
+  initialized @99 :Bool;
```
sshane, line 157: decide where initialized should go, probably carState!

Issue #208, opened 2021-10-16
msgq: Investigate using condition variables instead of signals


pd0wm:
Would probably require introducing mutexes around signaling readers, but would fix a rare lost wakeup in `msgq_poll`

PR #617, opened 2024-05-30
msgq: use shared memory mutex and condition variables for synchronization

deanlee:
This PR resolve issue #208, enhances the msgq  with shared memory mutexes and condition variables for improved IPC synchronization. While handling signals interrupts all processes, shared memory mechanisms provide a more robust and efficient solution, reducing overhead and complexity in high-frequency event scenarios.

key changes:

Each block of shared memory used for messages now includes a header:
```

struct SharedMemoryHeader {
  pthread_mutex_t mutex;
  pthread_cond_t cond;
  bool initialized;
};
```

This header contains the IPC mutex, condition variable, and an initialization flag.

When sending a message, the corresponding condition variable is used to notify the receiver to retrieve the message.
The Poller uses a global shared memory context, PollerContext, to handle message notifications and reception efficiently.

@pd0wm  If you have a moment, your review of this PR would be greatly appreciated to ensure I haven't missed anything or introduced new issues.

### DIFF
--- a/SConscript
+++ b/SConscript
@@ -15,7 +15,7 @@ msgq_objects = env.SharedObject([
   'msgq/msgq.cc',
 ])
 msgq = env.Library('msgq', msgq_objects)
-msgq_python = envCython.Program('msgq/ipc_pyx.so', 'msgq/ipc_pyx.pyx', LIBS=envCython["LIBS"]+[msgq, "zmq", common])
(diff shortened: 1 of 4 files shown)


deanlee:
Closed. linux `futex` appears to be a more efficient solution compared to using mutex and condition variables.

Issue #13, opened 2019-11-20
Debug_controls.py


vivekgr92:
I am not able to run the debug_control.py. when I run , I get this error

(openpilot) cyngn@vivekgr:~/roadtrain/openpilot/tools/carcontrols$ python debug_controls.py 
make: Nothing to be done for 'all'.
make: Nothing to be done for 'all'.
[ CC ] can_list_to_can_capnp.o
clang++ -std=c++11 -g -fPIC -I../ -I../../ -O2 -Werror=implicit-function-declaration -Werror=incompatible-pointer-types -Werror=int-conversion -Werror=return-type -Werror=format-extra-args -I/usr/include/libusb-1.0 -MMD \
          -Iinclude -I.. -I../.. \
          -I../../phonelibs/capnp-cpp/include \
          -I../../phonelibs/zmq/aarch64/include \
          -I../../selfdrive/messaging \
          -c -o 'can_list_to_can_capnp.o' 'can_list_to_can_capnp.cc'
can_list_to_can_capnp.cc:6:10: fatal error: 'cereal/gen/cpp/log.capnp.h' file not found
#include "cereal/gen/cpp/log.capnp.h"

Would be great if someone can help me.

vivekgr92:
Thanks , The solution is to make it again , so that gen file is created.

PR #624, opened 2024-06-26
Properly checking the return value of ipc_sendrecv_with_fds

deanlee:
This PR addresses an issue where the VIPC server closes the socket if the required stream type is unavailable:
https://github.com/commaai/msgq/blob/23cb05a80dec50a52f2e0fa7f7870ae920b3f2e8/msgq/visionipc/visionipc_server.cc#L136-L140
This causes the `ipc_sendrecv_with_fds` function in the VIPC client to return 0. However, the client does not properly handle this error and instead relies on two assertions that are always true:
https://github.com/commaai/msgq/blob/23cb05a80dec50a52f2e0fa7f7870ae920b3f2e8/msgq/visionipc/visionipc_client.cc#L56-L59

This PR introduces a proper check to ensure that the return value of `ipc_sendrecv_with_fds` is correctly handled if it is less than or equal to 0. These changes resolve the following bugs:

**1. Assertion in getAvailableStreams**

the vipc client still can connect to the ipc path even if the camerad is not running.   This could be due to the camerad not shutting down properly and the ipc path not being removed.

`ipc_sendrecv_with_fds` will returns -1 In this case. this causes the following assertion to fail.

> ui: cereal/visionipc/visionipc_client.cc:133: static std::set<VisionStreamType> VisionIpcClient::getAvailableStreams(const std::string &, bool): Assertion `(r >= 0) && (r % sizeof(VisionStreamType) == 0)' failed.

This bug can be reproduced by clicking “preview driver camera" button quickly and continuously in the UI interface.

**2. vipcClient:connect will returns true even if an attempt is made to connect to a non-existent stream type**
  the ipc_sendrecv_with_fds will return 0 if required stream type is not available. return false in this case.
 
This bug can be reproduced by run `watch3,` and run `replay --demo` . quit replay after the road camera displayed in watch3.  and run `replay --demo --ecam --dcam`. the watch3 will assert and quit:

> watch3: msgq_repo/msgq/visionipc/visionipc_client.cc:96: VisionBuf *VisionIpcClient::recv(VisionIpcBufExtra *, const int): Assertion `packet->idx < num_buffers' failed.

**3. crash in  VideoWidget::vipcAvailableStreamsUpdated:**

> #2  0x00007f882bce5729 in __assert_fail_base (fmt=0x7f882be7b588 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=0x5618e8 "", file=0x5617ff "repeat-1", line=263, function=<optimized out>) at assert.c:92
#3  0x00007f882bcf6fd6 in __GI___assert_fail (assertion=0x5618e8 "", file=0x5617ff "repeat-1", line=263, function=0x5618be "\311?") at assert.c:101
#4  0x000000000047ebce in QString::QString(char const*) (this=0x7fffde080230, ch=0x2 <error: Cannot access memory at address 0x2>) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qstring.h:700
#5  VideoWidget::vipcAvailableStreamsUpdated(std::set<VisionStreamType, std::less<VisionStreamType>, std::allocator<VisionStreamType> >) (this=Python Exception <class 'gdb.MemoryError'> Cannot access memory at address 0x0: 
    0x6546a8 <StreamNotifier::instance()::notifier+8>, streams=#6  0x000000000047e9a2 in VideoWidget::loopPlaybackClicked() (this=<optimized out>) at tools/cabana/videowidget.cc:198
#7  0x0000000001bb68e0 in  ()
#8  0x0000000001bb68b0 in  ()
#9  0x0000000001eb9680 in  ()
#10 0x0000000001c4ef10 in  ()
#11 0x00007fffde080240 in  ()
#12 0x00007f882cca5b40 in  () at /lib/x86_64-linux-gnu/libQt5Core.so.5
#13 0x00007f882ca10328 in QMetaObject::activate(QObject*, int, int, void**) () at /lib/x86_64-linux-gnu/libQt5Core.so.5

### DIFF
--- a/msgq/visionipc/visionipc_server.cc
+++ b/msgq/visionipc/visionipc_server.cc
@@ -134,7 +134,7 @@ void VisionIpcServer::listener(){
     }
 
     if (buffers.count(type) <= 0) {
-      std::cout << "got request for invalid buffer type: " << type << std::endl;
+      // std::cout << "got request for invalid buffer type: " << type << std::endl;
       close(fd);
       continue;
     }
--- a/msgq/visionipc/visionipc_client.cc
+++ b/msgq/visionipc/visionipc_client.cc
@@ -48,14 +48,22 @@ bool VisionIpcClient::connect(bool blocking){
   }
   // Send stream type to server to request FDs
   int r = ipc_sendrecv_with_fds(true, socket_fd, &type, sizeof(type), nullptr, 0, nullptr);
+  if (r <= 0) {
+    close(socket_fd);
+    return false;
+  }
+
   assert(r == sizeof(type));
 
   // Get FDs
   int fds[VISIONIPC_MAX_FDS];
   VisionBuf bufs[VISIONIPC_MAX_FDS];
   r = ipc_sendrecv_with_fds(false, socket_fd, &bufs, sizeof(bufs), fds, VISIONIPC_MAX_FDS, &num_buffers);
+  if (r <= 0) {
+    close(socket_fd);
+    return false;
+  }
 
-  assert(num_buffers >= 0);
   assert(r == sizeof(VisionBuf) * num_buffers);
 
   // Import buffers
@@ -122,11 +130,19 @@ std::set<VisionStreamType> VisionIpcClient::getAvailableStreams(const std::strin
   // Send VISION_STREAM_MAX to server to request available streams
   int request = VISION_STREAM_MAX;
   int r = ipc_sendrecv_with_fds(true, socket_fd, &request, sizeof(request), nullptr, 0, nullptr);
-  assert(r == sizeof(request));
+  if (r <= 0) {
+    close(socket_fd);
+    return {};
+  }
 
   VisionStreamType available_streams[VISION_STREAM_MAX] = {};
   r = ipc_sendrecv_with_fds(false, socket_fd, &available_streams, sizeof(available_streams), nullptr, 0, nullptr);
-  assert((r >= 0) && (r % sizeof(VisionStreamType) == 0));
+  if (r <= 0) {
+    close(socket_fd);
+    return {};
+  }
+
+  assert((r % sizeof(VisionStreamType)) == 0);
   close(socket_fd);
   return std::set<VisionStreamType>(available_streams, available_streams + r / sizeof(VisionStreamType));
 }


msgq/visionipc/visionipc_server.cc
```diff
@@ -134,7 +134,7 @@ void VisionIpcServer::listener(){
     }
 
     if (buffers.count(type) <= 0) {
-      std::cout << "got request for invalid buffer type: " << type << std::endl;
+      // std::cout << "got request for invalid buffer type: " << type << std::endl;
```
deanlee, line 137: After this PR, this message will spam the console.

sshane:
@deanlee should we check for the specific errno in this case so we can still catch unexpected errors? (ECONNRESET)

sshane:
Looks like there's another crash spamming driver view even with this PR:

```
(gdb) bt
#0  0x0000007fb4d40de0 in main_arena () from /lib/aarch64-linux-gnu/libc.so.6
#1  0x0000007fb3c16624 in ?? () from /lib/aarch64-linux-gnu/libffi.so.8
#2  0x0000007fb3c1383c in ?? () from /lib/aarch64-linux-gnu/libffi.so.8
#3  0x0000007fb50567c4 in ?? () from /lib/aarch64-linux-gnu/libwayland-client.so.0
#4  0x0000007fb505711c in ?? () from /lib/aarch64-linux-gnu/libwayland-client.so.0
#5  0x0000007fb5057444 in wl_display_dispatch_queue_pending () from /lib/aarch64-linux-gnu/libwayland-client.so.0
#6  0x0000007fafe36078 in EglWaylandWlWindowSurface::CommitBuffer(void*, unsigned int) () from /lib/aarch64-linux-gnu/libeglSubDriverWayland.so
#7  0x0000007fafe37e78 in EglWaylandUpdater::UpdaterThread() () from /lib/aarch64-linux-gnu/libeglSubDriverWayland.so
#8  0x0000007fb4c1597c in start_thread (arg=0x0) at ./nptl/pthread_create.c:447
#9  0x0000007fb4c7b7dc in thread_start () at ../sysdeps/unix/sysv/linux/aarch64/clone.S:79
```

PR #625, opened 2024-07-04
use futex for IPC notifications

deanlee:
This PR replaces the use of `std::signal `with `futex` for inter-process notifications between publishers and subscribers. improving the robustness and performance of IPC notifications.  Tested on PC and c3x device.

Benefits:

1.  Futex provides a cleaner and more reliable synchronization mechanism than signals. Signals can interrupt system calls across all subscriber processes, adding complexity and impacting overall system performance.
2.  Futex minimizes context switches and eliminates signal handler overhead, thereby reducing synchronization latency. It primarily operates in user space and enters kernel space only when necessary, unlike std::signal, which always involves kernel space operations.
3. Futex provides a more reliable and easier-to-debug synchronization mechanism than signals, ensuring robust inter-process communication.

### DIFF
--- a/SConscript
+++ b/SConscript
@@ -12,6 +12,7 @@ msgq_objects = env.SharedObject([
   'msgq/impl_zmq.cc',
   'msgq/impl_msgq.cc',
   'msgq/impl_fake.cc',
+  'msgq/futex.cc',
   'msgq/msgq.cc',
 ])
 msgq = env.Library('msgq', msgq_objects)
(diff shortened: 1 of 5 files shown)


adeebshihadeh:
This is a pretty core change to msgq. I'd like to do it, but we need to make sure that it hits all of our goals. Fixing the `EINTR` spam is great, but let's also make sure it'll work on macOS.

PR #626, opened 2024-07-09
Convert tests to pytests

UkuLoskit:
Part of https://github.com/commaai/openpilot/issues/32941


maxime-desroches:
msgq is horrible to work in, we have to clean this up

PR #634, opened 2024-11-16
Refactor get_port to use FNV-1a hash for consistency across platforms

deanlee:
resolve https://github.com/commaai/openpilot/issues/34031 :

>  the issue is caused by a port mismatch between the Mac and the device:
> 
>     Attempting to connect to zmq endpoint: tcp://172.16.0.20:60942
>     Expected port on the device: 63557
> 
> It seems the static int get_port(std::string endpoint) function returns different port values on macOS and Linux, likely due to platform differences in how std::hash behaves.

This pr fixes the issue where the `get_port` function returns different port values on macOS and Linux due to platform-specific behavior of `std::hash`. The mismatch in port values causes connection failures in ZMQ. The solution replaces` std::hash `with a custom FNV-1a hash function to ensure consistent port values across platforms, and reducing the risk of collisions.

### DIFF
--- a/msgq/impl_zmq.cc
+++ b/msgq/impl_zmq.cc
@@ -7,10 +7,19 @@
 
 #include "msgq/impl_zmq.h"
 
+static size_t fnv1a_hash(const std::string &str) {
+    const size_t fnv_prime = 0x100000001b3;
+    size_t hash_value = 0xcbf29ce484222325;
+    for (char c : str) {
+        hash_value ^= (unsigned char)c;
+        hash_value *= fnv_prime;
+    }
+    return hash_value;
+}
+
 //FIXME: This is a hack to get the port number from the socket name, might have collisions
 static int get_port(std::string endpoint) {
-    std::hash<std::string> hasher;
-    size_t hash_value = hasher(endpoint);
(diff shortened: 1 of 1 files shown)


pbassut:
Confirmed this works! 
Thank you @deanlee for such quick turn around on this one

sshane:
@deanlee this breaks PlotJuggler, can you make a PR to fix?

adeebshihadeh:
Just need to rebuild PJ?

devtekve:
Thanks @deanlee, this makes it easier to derive the same port from other languages!

pbassut:
yea, confirmed PJ is broken. I wasn't even able to build/run PJ before so unsure if it was caused by this PR.

devtekve:
I can't stream either, but I checked with older commit and I still couldn't stream either. So I don't think it was this change.

pbassut:
Can we confirm that PJ only needs a rebuild for this to be fixed?

deanlee:
> Can we confirm that PJ only needs a rebuild for this to be fixed?

y.  make sure to update to the latest master branch of common/PlotJugglere before rebuilding.

PR #635, opened 2024-11-24
macos: use /tmp instead of /dev/shm

devtekve:
In macos a lot of the tests in OP still seem to rely on "/dev/shm" so this, alongside with https://github.com/commaai/openpilot/pull/34097 should unblock those cases

### DIFF
--- a/msgq/msgq_tests.cc
+++ b/msgq/msgq_tests.cc
@@ -1,6 +1,13 @@
 #include "catch2/catch.hpp"
 #include "msgq/msgq.h"
 
+// Define queue path based on platform
+#ifdef __APPLE__
+  #define SHM_PATH "/tmp/test_queue"
+#else
+  #define SHM_PATH "/dev/shm/test_queue"
+#endif
+
 TEST_CASE("ALIGN")
 {
   REQUIRE(ALIGN(0) == 0);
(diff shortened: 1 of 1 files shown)


adeebshihadeh (changes requested):
Can you pull this out into a constant to reduce all the ifdef everywhere?

devtekve:
I will , sorry ! I’ve been busy with MADS rewrite in SP

msgq/msgq_tests.cc
```diff
@@ -383,15 +390,15 @@ TEST_CASE("1 publisher, 1 slow subscriber", "[integration]")
       msgq_msg_close(&msg1);
     }
   }
-
+
```
adeebshihadeh, line 393: revert this

msgq/event.cc
```diff
@@ -21,7 +21,11 @@
 void event_state_shm_mmap(std::string endpoint, std::string identifier, char **shm_mem, std::string *shm_path) {
   const char* op_prefix = std::getenv("OPENPILOT_PREFIX");
 
-  std::string full_path = "/dev/shm/";
+  #ifdef __APPLE__
+    std::string full_path = "/tmp/";
+  #else
+    std::string full_path = "/dev/shm/";
+  #endif
```
adeebshihadeh, line 28: this should be pulled out

msgq/msgq.cc
```diff
@@ -87,7 +87,11 @@ int msgq_new_queue(msgq_queue_t * q, const char * path, size_t size){
   assert(size < 0xFFFFFFFF); // Buffer must be smaller than 2^32 bytes
   std::signal(SIGUSR2, sigusr2_handler);
 
-  std::string full_path = "/dev/shm/";
+  #ifdef __APPLE__
+    std::string full_path = "/tmp/";
+  #else
+    std::string full_path = "/dev/shm/";
```
adeebshihadeh, line 93: same here

adeebshihadeh (changes requested):

adeebshihadeh:
`/dev/shm` and `/tmp` should only be defined once

PR #636, opened 2024-11-28
msgq: fix permission of subscribers

katsuster:
The current implementation of MSGQ and VisionIPC allows subscriber processes to potentially overwrite publisher's data because subscribers have write permissions for shared data area of the queue.

This patch changes mmap() protection of subscribers from `PROT_READ | PROT_WRITE` to `PROT_READ` only.
This prevents unintended writes by subscribers and ensures the integrity of publisher's data.

### DIFF
--- a/msgq/msgq_tests.cc
+++ b/msgq/msgq_tests.cc
@@ -1,3 +1,7 @@
+#include <signal.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+
 #include "catch2/catch.hpp"
 #include "msgq/msgq.h"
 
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Seems like it's still easy to setup a subscriber that can write. Can you add a test to ensure that doesn't happen? Or even better, can you think of a way to rewrite it to make that bug less likely?

katsuster:
Thank you for your comment!
It's OK that subscribers can setup write permission easily. I don't intend to ensure protecting from 'malicious subscribers' that  try to get write permissions and destroy publisher's data intentionally.

I'm targeting for 'normal subscribers' that use `PubMaster/SubMaster` API cannot take unneeded write permissions.
I also want to prevent silent data corruption by unintentional write to shared area due to their implementation bugs.

katsuster:
@adeebshihadeh 
Sorry for late reply.
I introduced a new test for detecting SIGSEGV. Please review it.

katsuster:
@adeebshihadeh 
Sorry for interrupt, How about this? Is there anything to improve?

PR #638, opened 2024-12-17
VisionIpc: fix exception if server is killed while client is getting streams

sshane:
Fixes a ui crash in openpilot when ignition is rapidly cycled from off to on to off. The first assertion always passes (due to timing?), but the second can sometimes return -1.

This return code comes from this line where the errno is always 104 (ECONNRESET):

https://github.com/commaai/msgq/blob/434ed2312c980b5504a4a75001d04c3ecbddf93f/msgq/visionipc/visionipc.cc#L92-L93


sshane:
Fix already open: https://github.com/commaai/msgq/pull/624

PR #641, opened 2025-01-18
remove receive signal handler

sshane:
The only difference when receiving a termination signal is this line not being ran, but the program would be exiting anyway if this is the case:

```
msgq_msg_close(&msg); // Free unused message on exit
```

With the handler, the polling is not interrupted, so we hang until the timeout is up. If this is set high, we can get stuck here on exit.

### DIFF
--- a/msgq/impl_msgq.cc
+++ b/msgq/impl_msgq.cc
@@ -2,19 +2,9 @@
 #include <cstring>
 #include <iostream>
 #include <cstdlib>
-#include <csignal>
 #include <cerrno>
 
 #include "msgq/impl_msgq.h"
-
-
-volatile sig_atomic_t msgq_do_exit = 0;
-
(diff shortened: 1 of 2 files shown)


deanlee:
I like this solution; it resolves the issue while reducing complexity.

sshane:
Something about this is causing pandad to hang, will need to check why

deanlee:
It may not be directly related to the pandad issue, but we should removing the` if (errno == EINTR)` check in pandad after this PR.
```c++

  // run as fast as messages come in
  while (!do_exit && check_all_connected(pandas)) {
    std::unique_ptr<Message> msg(subscriber->receive());
    if (!msg) {
      if (errno ==   EINTR) {
        do_exit = true;
      }
      continue;
    }
```

sshane:
thanks, that was it!

